PR #19, opened 2022-06-16
Ability to dual publish

samruddhikhandale:
Publish the same image to `devcontainers` and `vscode/devcontainers` by retagging


Chuxel:
It might make sense to refer to the second spot just as "secondary registry" rather than calling out VS Code by name. e.g., we could at some point switch to some other container registry for publishing. Might make sense to also make it optional. Otherwise it looks like this would work!

joshspicer (approved):
+1 on the secondary registry nit, but otherwise lgtm!

edgonmsft (approved):
Yep!

PR #56, opened 2022-08-03
Fix bug: Action 'Update version history' gets skipped for 'push'

samruddhikhandale:
Similar to https://github.com/devcontainers/images/pull/55

Fixes: https://github.com/devcontainers/images/runs/7659461741?check_suite_focus=true


Stas0089:
97bae6b26aa7ea34fe9a38b90f9eb919d2f18895

PR #80, opened 2022-08-25
Rename src folders and references to match with image names

samruddhikhandale:
- Renames src folders (and corresponding references) of the folders whose image name did not match with folder name 
    - alpine --> base-alpine 
    - debian --> base-debian 
    - ubuntu --> base-ubuntu 
    - codespaces --> universal
- README changes for universal image (de-branding)
- (Other nit) Renamed `imageVersion` in `manifest.json` to `version`


build/src/push.js
```diff
@@ -122,9 +122,9 @@ async function pushImage(definitionId, repo, release, updateLatest,
             if (replaceImage || !await isDefinitionVersionAlreadyPublished(definitionId, release, registry, registryPath, variant)) {
 
                 let platformParams = "";
-                // Codespaces image does not need to be multi-arch
-                // ubuntu:focal image supports multiarch but codespaces doesn't. Hence, the build fails similar to https://github.com/docker/buildx/issues/235
-                if (definitionId != "codespaces") {
+                // Universal image does not need to be multi-arch
+                // ubuntu:focal image supports multiarch but Universal does not. Hence, the build fails similar to https://github.com/docker/buildx/issues/235
+                if (definitionId != "universal") {
```
samruddhikhandale, line 127: TODO: I'll create a follow up cleanup PR which renames `definition` ref to `image`  

> if (imageId != "universal") {

src/universal/README.md
```diff
@@ -24,41 +24,15 @@ If you use GitHub Codespaces, this is the "universal" image that is used by defa
...
+You can decide how often you want updates by referencing a [semantic version](https://semver.org/) of each image.
+For example:
 
 - `mcr.microsoft.com/devcontainers/universal:2-focal`
 - `mcr.microsoft.com/devcontainers/universal:2.0-focal`
 - `mcr.microsoft.com/devcontainers/universal:2.0.6-focal`
 
 See [history](history) for information on the contents of each version and [here for a complete list of available tags](https://mcr.microsoft.com/v2/devcontainers/universal/tags/list).
 
-## Accessing the container using SSH, SCP, or SSHFS
```
jkeech, line 35: Not sure if it's still worth calling out that this image includes a running SSH server. We could still keep part of these docs if folks find it useful.

jkeech (approved):
Looks good. Approved with one comment.

src/universal/README.md
Chuxel: Yeah given this mentions GitHub Codespaces and Remote - Containers we should probably remove it.  This pre-dates the GH CLI integration for Codespaces. The SSHD feature has this info for those that add it that way https://github.com/devcontainers/features/tree/main/src/sshd. 

I suppose we could mention that the SSH feature is already added to the image and link to its README for that matter.

samruddhikhandale: > I suppose we could mention that the SSH feature is already added to the image and link to its README for that matter.

Makes sense. Added !

src/base-debian/manifest.json
```diff
@@ -1,10 +1,10 @@
 {
+	"version": "0.202.8",
 	"variants": [
 		"buster",
 		"bullseye",
 		"stretch"
 	],
-	"imageVersion": "0.202.8",
```
joshspicer, line 7: Was this a typo from before?

joshspicer: never mind :)

joshspicer (approved):

PR #82, opened 2022-08-26
Universal: update manifest

samruddhikhandale:
Updates manifest with ML/python libraries

### DIFF
--- a/src/universal/manifest.json
+++ b/src/universal/manifest.json
@@ -107,6 +107,19 @@
 			"github.com/go-delve/delve": null,
 			"github.com/golangci/golangci-lint": null
 		},
+		"pip": [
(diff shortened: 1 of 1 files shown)


src/universal/manifest.json
```diff
@@ -107,6 +107,19 @@
 			"github.com/go-delve/delve": null,
 			"github.com/golangci/golangci-lint": null
 		},
+		"pip": [
```
joshspicer, line 110: This is not autogenerated, correct?

joshspicer (approved):

src/universal/manifest.json
samruddhikhandale: Not autogenerated.

PR #85, opened 2022-08-26
Action - Fix bug for update version history

samruddhikhandale:
Fixes - https://github.com/devcontainers/images/runs/8045107992?check_suite_focus=true

image


joshspicer (approved):

joshspicer:
Never seen this part of the tool - interesting!

PR #88, opened 2022-08-30
fix bug - update image history

samruddhikhandale:
Due to incorrect pagination, history files were not created for `"javascript-node"` and `"typescript-node"`. Fixing that! 

image

Expected -

image


build/src/utils/config.js
```diff
@@ -392,7 +392,8 @@ function getSortedDefinitionBuildList(page, pageTotal, definitionsToSkip) {
     if (allPages.length > pageTotal) {
         // If too many pages, add extra pages to last one
         console.log(`(!) Not enough pages to for target page size. Adding excess definitions to last page.`);
-        for (let i = pageTotal; i < allPages.length; i++) {
```
joshspicer, line 395: Seems strange to me this change would be necessary, given it used to work in vscode-dev-containers 🤔 

That said, lgtm

joshspicer (approved):

PR #92, opened 2022-08-30
fix bug: incorrect 'latest' tag for base image

samruddhikhandale:


build/src/push.js
```diff
@@ -132,11 +132,6 @@ async function pushImage(definitionId, repo, release, updateLatest,
                 const workingDir = path.resolve(dotDevContainerPath, context);
                 const imageNameParams = imageNamesWithVersionTags.reduce((prev, current) => prev.concat(['--image-name', current]), []);
 
-                // Do not build and push the "latest" tag when pushing the "dev" images
```
samruddhikhandale, line 135: This was something I had added while switching building and pushing images with cli

samruddhikhandale: `imageNamesWithVersionTags` contains the correct tags.
It does have `:latest` when it's actually needed. No need of a special condition

joshspicer (approved):

PR #99, opened 2022-09-02
fix - update history bug

samruddhikhandale:
Fixes - https://github.com/devcontainers/images/runs/8147784328?check_suite_focus=true

image

`jupyter-lab` is found within a container. One of `~/.bashrc, ~/.bash_profile or ~/.profile` updates the `PATH`, these are not sourced when running commands only with -c.

### DIFF
--- a/src/universal/manifest.json
+++ b/src/universal/manifest.json
@@ -200,7 +200,7 @@
 			},
 			"Jupyter Lab": {
 				"cgIgnore": true,
-				"versionCommand": "jupyter-lab --version",
+				"versionCommand": "export PATH=$PATH && jupyter-lab --version",
 				"path": "/home/codespace/.local/bin/jupyter-lab",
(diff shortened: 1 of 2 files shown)


src/universal/manifest.json
```diff
@@ -200,7 +200,7 @@
 			},
 			"Jupyter Lab": {
 				"cgIgnore": true,
-				"versionCommand": "jupyter-lab --version",
+				"versionCommand": "export PATH=$PATH && jupyter-lab --version",
```
joshspicer, line 203: Alternatively you could supply the full path to where the jypyter-lab executable is supposed to be installed.

samruddhikhandale: Umm (unless I'm understanding). that's done by the `Python` Feature within the `universal` image which isn't reflecting with `bash -c`

joshspicer (approved):

PR #103, opened 2022-09-06
universal - Add CMD and containerUser

samruddhikhandale:
- Copies over missing CMD from [here](https://github.com/microsoft/vscode-dev-containers/blob/main/containers/codespaces-linux/.devcontainer/base.Dockerfile#L126)

- The new image has config.user: "root", however the old one has `codespace` (set [here](https://github.com/microsoft/vscode-dev-containers/blob/main/containers/codespaces-linux/.devcontainer/base.Dockerfile#L136)). Fixing that.


edgonmsft (approved):
LGTM

PR #139, opened 2022-09-29
Filter unwanted Jupyter kernels in Universal image

jungaretti:
Adds the `jupyter.kernels.filter` VS Code setting to the Universal image to filter unwanted Python kernels. This setting makes for a much cleaner VS Code experience when using the Jupyter extension.

This won't filter new Conda environments or Python installations that users add. It just hides the extra installations in the universal image that don't include our Jupyter and ML tooling.

Before:
Many Python installations in the kernel picker

Note that the `Suggested` kernel isn't our primary installation of Python and that there are several entries for our primary installation (because of our symlinks).

After:
Only our primary Python installation in the kernel picker

The only kernel that appears in the picker is our primary installation of Python. Its path matches the output of `which python3`.

### DIFF
--- a/src/universal/.devcontainer/devcontainer.json
+++ b/src/universal/.devcontainer/devcontainer.json
@@ -112,16 +112,16 @@
                 "go.useLanguageServer": true,
                 "go.gopath": "/go",
                 "python.defaultInterpreterPath": "/home/codespace/.python/current/bin/python3",
-                "python.linting.enabled": true,
-                "python.linting.pylintEnabled": true,
                 "python.formatting.autopep8Path": "/usr/local/py-utils/bin/autopep8",
                 "python.formatting.blackPath": "/usr/local/py-utils/bin/black",
                 "python.formatting.yapfPath": "/usr/local/py-utils/bin/yapf",
                 "python.linting.banditPath": "/usr/local/py-utils/bin/bandit",
+                "python.linting.enabled": true,
                 "python.linting.flake8Path": "/usr/local/py-utils/bin/flake8",
                 "python.linting.mypyPath": "/usr/local/py-utils/bin/mypy",
                 "python.linting.pycodestylePath": "/usr/local/py-utils/bin/pycodestyle",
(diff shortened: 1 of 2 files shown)


samruddhikhandale (commented):
🚀  Looks great to me! 

@jungaretti Could you help bump the patch version in https://github.com/devcontainers/images/blob/main/src/universal/manifest.json#L2 so that we are ready to publish the `universal` image with this change?

src/universal/.devcontainer/devcontainer.json
```diff
@@ -111,7 +111,7 @@
                 "go.toolsManagement.checkForUpdates": "local",
                 "go.useLanguageServer": true,
                 "go.gopath": "/usr/local/go/bin/go",
```
samruddhikhandale, line 113:
```suggestion
                "go.gopath": "/go",
```

Could you help fix this when you are at it? 😄 
We have been going back and forth with `GOPath` variable, but seems like this should be `/go` - syncing with the GO Feature. 
Reference - https://github.com/devcontainers/features/blob/main/src/go/devcontainer-feature.json#L30

src/universal/.devcontainer/devcontainer.json
```diff
@@ -123,6 +123,36 @@
...
+                        "type": "pythonEnvironment"
+                    },
+                    {
+                        "path": "/bin/python3",
+                        "type": "pythonEnvironment"
+                    }
+                ],
                 "lldb.executable": "/usr/bin/lldb",
                 "files.watcherExclude": {
                     "**/target/**": true
```
samruddhikhandale, line 158: @Chuxel  could you help me understand the scenario wherein we need to exclude this? 🤔 
Not sure if we should clean this up if it's no longer needed.

samruddhikhandale (approved):

src/universal/.devcontainer/devcontainer.json
jungaretti: https://github.com/devcontainers/images/pull/141

joshspicer (approved):

PR #140, opened 2022-09-30
Pin Python versions in Universal image

jungaretti:
Pins the Python versions in the Universal image at 3.10.4 and 3.9.7 to ensure compatibility with Oryx.

https://github.com/microsoft/Oryx/blob/main/doc/supportedPlatformVersions.md#python

Python 3.10.4


joshspicer (approved):
nice clean PR thanks to Features 🚀

PR #141, opened 2022-09-30
Remove files.watcherExclude from Universal image

jungaretti:
Removes the `files.watcherExclude` VS Code settings from the Universal image.

### DIFF
--- a/src/universal/manifest.json
+++ b/src/universal/manifest.json
@@ -1,5 +1,5 @@
 {
-	"version": "2.0.17",
+	"version": "2.0.18",
 	"build": {
 		"latest": true,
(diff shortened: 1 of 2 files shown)


src/universal/manifest.json
```diff
@@ -1,5 +1,5 @@
 {
-	"version": "2.0.19",
```
samruddhikhandale, line 2: Can we change it to `2.0.18` ? 
We are in a process of publishing `prod` tags for `universal`, we could even have this PR change join the release.

jungaretti: Fixed

samruddhikhandale (approved):
LGTM!

PR #162, opened 2022-10-18
Adds git Feature to images

samruddhikhandale:
- Adds `git` Feature to all images except `alpine` and `universal`
- Modifies `common-alpine` to temporarily install git from source


src/base-alpine/.devcontainer/library-scripts/common-alpine.sh
```diff
@@ -95,7 +95,23 @@ if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
         which \
         ncdu \
         shadow \
-        strace
+        strace \
```
joshspicer, line 98: Where did these come from?

samruddhikhandale: These packages are required to build git from source. Without them, the `make` command fails

joshspicer (approved):
yay for Features!

PR #176, opened 2022-10-25
Python version 3.11

edgonmsft:


src/python/manifest.json
```diff
@@ -56,12 +56,15 @@
 			"python:${VERSION}-${VARIANT}"
 		],
 		"variantTags": {
-			"3.10-bullseye": [
+			"3.11-bullseye": [
 				"python:${VERSION}-3.10",
```
Chuxel, line 60: Should be:

```suggestion
				"python:${VERSION}-3.11",
```

src/python/manifest.json
```diff
@@ -74,7 +77,7 @@
 			"3.6-bullseye": [
 				"python:${VERSION}-3.6"
 			],
```
Chuxel, line 79: You can remove this since 3.6 is out of support now and won't be getting security patches.

src/python/manifest.json
```diff
@@ -1,19 +1,19 @@
...
 		"3.8-buster",
-		"3.7-buster",
-		"3.6-buster"
+		"3.7-buster"
 	],
 	"build": {
-		"latest": "3.10-bullseye",
+		"latest": "3.11-bullseye",
 		"rootDistro": "debian",
 		"architectures": {
```
Chuxel, line 18: You'll need to add a 3.11-bullseye and 3.11-buster section here

You can drop the 3.6-bullseye/3.6-buster sections.

Chuxel:
The README will need to be updated as well to drop 3.6 and add 3.11 in a number of spots.

edgonmsft:
> The README will need to be updated as well to drop 3.6 and add 3.11 in a number of spots.
> 
> The Python template and feature then also need to be updated to remove 3.6 and add 3.11 to the list.

I've updated the feature already, I'll update the template.

src/python/manifest.json
```diff
@@ -1,21 +1,25 @@
 {
 	"version": "0.203.12",
```
Chuxel, line 2: You'll need to bump the version number here. The version check is for the whole image set rather than specific variants as things are today.

Chuxel (dismissed):
Other than bumping the image version that should do it!

EDIT: Actually looks like  x.x.13 is already bumped given https://github.com/devcontainers/images/pull/174 is not merged

jkeech (dismissed):

jkeech (approved):

PR #178, opened 2022-10-26
universal - fix oryx bug for Python projects

samruddhikhandale:
The `oryx` tool fails to automatically install python packages from `requirements.txt`. Fixing that.


src/universal/.devcontainer/local-features/setup-user/install.sh
```diff
@@ -38,6 +41,7 @@ ln -snf /usr/local/php/current $PHP_PATH
 PYTHON_PATH="/home/codespace/.python/current"
 mkdir -p /home/codespace/.python
 ln -snf /usr/local/python/current $PYTHON_PATH
+ln -snf /usr/local/python /opt/python
```
jungaretti, line 44: We may need to add `/opt/python` to the list of directories that the Jupyter extension filters, testing locally now...

https://github.com/devcontainers/images/blob/c7d4fc1862a0e46934c8f5c952c643b652d8bb4c/src/universal/.devcontainer/devcontainer.json#L151-L154

jungaretti: Before `ln -snf /usr/local/python /opt/python`:

Kernels filtered properly

After `ln -snf /usr/local/python /opt/python` and window reload:

Kernels still filtered properly

jungaretti: Looks good ✅

jungaretti (approved):
LGTM, thanks

jungaretti (approved):
Makes sense

edgonmsft (approved):

src/universal/test-project/test.sh
```diff
@@ -37,7 +37,7 @@ check "pydocstyle" pydocstyle --version
 check "bandit" bandit --version
 check "virtualenv" virtualenv --version
 count=$(ls /usr/local/python | wc -l)
-expectedCount=3 # 2 version folders + 1 current folder which links to either one of the version
+expectedCount=4 # 2 version folders + 1 current folder which links to either one of the version + 1 folder generated by oryx build command
```
samruddhikhandale, line 40: As we are now adding python files to `test-project` [in this PR ](https://github.com/devcontainers/features/pull/239). The `oryx build` tool is now detecting the pyton platfrom and trying to install `3.8.12`. Hence, updated this.

Before
image

After

image

samruddhikhandale: Need to be updated to fix this test

image

Issue #170, opened 2022-10-22
`mcr.microsoft.com/devcontainers/base:ubuntu` is still focal?


eitsupi:
The README says it is the latest LTS version, but it doesn't seem to be the case.

https://github.com/devcontainers/images/blob/35c84eded76ba0fb5ac6635c066022ab381e755e/src/base-ubuntu/README.md?plain=1#L24

Note that `docker.io/library/ubuntu:latest` has already switched to jammy right after jammy was released half a year ago.

bamurtaugh:
Thanks @eitsupi!

@samruddhikhandale I see this too - should we switch to Jammy, or update the readme?

samruddhikhandale:
Created PR - https://github.com/devcontainers/images/pull/179 which switches to `jammy`.

Thanks @eitsupi for catching this! ✨

samruddhikhandale:
Released `dev` tags which switches `base:dev-ubuntu` to `jammy`, looks good. I will release prod tags now. 

![image](https://user-images.githubusercontent.com/24955913/198339721-9e8b4842-d7fd-4f19-b7df-6f0681783d4f.png)

// cc @Chuxel

samruddhikhandale:
`https://github.com/devcontainers/images/pull/179` is live now which uses `jammy` 🚀

PR #179, opened 2022-10-26
base:ubuntu is switched to use `jammy`

samruddhikhandale:
Addresses - https://github.com/devcontainers/images/issues/170

// cc @Chuxel

### DIFF
--- a/src/base-ubuntu/manifest.json
+++ b/src/base-ubuntu/manifest.json
@@ -27,12 +27,12 @@
 		"variantTags": {
 			"jammy": [
 				"base:${VERSION}-ubuntu-22.04",
-				"base:${VERSION}-ubuntu22.04"
+				"base:${VERSION}-ubuntu22.04",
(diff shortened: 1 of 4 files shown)


src/base-ubuntu/manifest.json
```diff
@@ -1,5 +1,5 @@
 {
-	"version": "0.203.9",
+	"version": "0.204.0",
```
samruddhikhandale, line 2: @Chuxel Do we bump a major version in this case? 😕 
Potentially this could be a breaking change.

Chuxel (commented):
You'll also need to update `"variantTags"` since that's where the `ubuntu` tag is actually assigned.

src/base-ubuntu/manifest.json
Chuxel: Yeah makes sense.

edgonmsft (approved):

PR #180, opened 2022-10-26
Add full-flow Oryx text for Python

jungaretti:
Adds a full-flow Oryx text for Python using a sample project.


src/universal/test-project/test.sh
```diff
@@ -134,6 +134,10 @@ check "run-puppeteer" node puppeteer.js
 # Check Oryx
 check "oryx" oryx --version
 
+# Make sure that Oryx builds Python projects correctly
+check "oryx-build-python" oryx build --property python_version='3.10.4' --property packagedir='/home/codespace/.local/lib/python3.10/site-packages' ./sample/python
```
samruddhikhandale, line 138: Even my first instance was to write a similar test.
But now that I am thinking, should  we also write a test to repro the `universal` images' post create command.

https://github.com/devcontainers/images/blob/main/src/universal/.devcontainer/devcontainer.json#L104

This particular command is something that is executed by oryx

src/universal/test-project/test.sh
```diff
@@ -134,6 +134,10 @@ check "run-puppeteer" node puppeteer.js
...
+check "oryx-build-python-result" python3 ./sample/python/src/solve.py
```
samruddhikhandale, line 139: Nice test. 

💡 We could also test if oryx worked with `pip list | grep mpmath`

src/universal/test-project/test.sh
samruddhikhandale, line 138:
```suggestion
pythonVersion=$(python -V 2>&1 | grep -Po '(?<=Python )(.+)')
pythonSite=`python -m site --user-site`
check "oryx-build-python" oryx build --property python_version="${pythonVersion}" --property packagedir="${pythonSite}" ./sample/python
```

src/universal/test-project/test.sh
jungaretti: Good idea, I'll add that

src/universal/test-project/test.sh
samruddhikhandale: Not needed. Discussed offline

samruddhikhandale (approved):
Thanks! ✨  The smoke test for `universal` is supposed to fail.

It is fixed in https://github.com/devcontainers/images/pull/178

Musti331:
₺180

Musti331:
#180

Musti331:
@@ -134,6 +134,10 @@

PR #184, opened 2022-10-27
Build & push actions - Configure  jobs with variants

samruddhikhandale:
Attached logs for sample jobs, which shows  how the new logic for pagination works.

> Command - build/vscdc push --no-push --registry mcr.microsoft.com --registry-path devcontainers --release v1.3 --page 57  --pageTotal 57 --replace true

[page1-logs.txt](https://github.com/devcontainers/images/files/9883679/page1-logs.txt)
[page30-logs.txt](https://github.com/devcontainers/images/files/9883680/page30-logs.txt)
[page57-logs.txt](https://github.com/devcontainers/images/files/9883681/page57-logs.txt)


build/src/push.js
```diff
@@ -41,17 +41,26 @@ async function push(repo, release, updateLatest, registry, registryPath, stubReg
...
-            currentDefinitionId, repo, release, updateLatest, registry, registryPath, stubRegistry, stubRegistryPath, prepOnly, pushImages, replaceImages, secondaryRegistryPath);
-    });
+    if (definitionId) {
+        const variants = configUtils.getVariants(definitionId) || [null];
+        await asyncUtils.forEach(variants, async (variant) => {
+            console.log(`**** Pushing ${definitionId}: ${variant} ${release} ****`);
+            await pushImage(
+                definitionId, variant, repo, release, updateLatest, registry, registryPath, stubRegistry, stubRegistryPath, prepOnly, pushImages, replaceImages, secondaryRegistryPath);
+        });
+    } else {
```
joshspicer, line 51: When would we not have a `definitionId`?

joshspicer (approved):

build/src/push.js
samruddhikhandale: For dev and prod actions, we don't pass in `definition` as we want to build all the images.
For [re-push](https://github.com/devcontainers/images/blob/main/.github/workflows/push-again.yml#L82) image, we pass that when we'd like to build a single image

PR #191, opened 2022-11-02
Update universal - Remove 'tensorflow'

samruddhikhandale:
Remove following python packages from `universal` image.

 - `tensorflow`: The `universal` image currently does not support CUDA and `tensorflow` doesn't work without it. Hence, removing it to free `~ 1.4 GB` of image space.
 - `keras` - Has a hard dependency on `tensorflow` and it is just a wrapper for it.


jkeech (approved):

jungaretti (approved):
LGTM, thanks for cleaning this up

Issue #200, opened 2022-11-09
Update dotnet container image


timheuer:
.NET 7 just released.

Update container images for .NET 7 as default

Chuxel:
@timheuer Several PRs are queued up, but currently `universal` (aka "kitchen sink") has .NET 3 and 6.  I'm assuming this should now be 6 and 7?

timheuer:
Should be 3.1, 6, 7

-th
________________________________

Chuxel:
@timheuer Given 3.1 is EOL in Dec, why include it? Given the size of the image we are generally including only latest LTS for most languages by default.

timheuer:
If the update happens after EOL that makes sense but the reality is there are two LTS right now

-th
________________________________

Chuxel:
Yeah for the separate images we can do all three but I will have to defer to the Codespaces folks on the universal image given the desire there is to not grow it beyond its current size since that reduces usable storage.

Chuxel:
`mcr.microsoft.com/devcontainers/dotnet:7.0` and variants are now up. The dotnet Dev Container Feature is now also updated.

@samruddhikhandale is working on the template https://github.com/devcontainers/templates/pull/39 which can get merged now that the images are released.

The `unviersal` and Codespace's use of it by default requires testing, though that is also in flight at this point w/.NET 6 and 7 being there.  (3.1, 6.0, and 7.0 are al there for the dotnet image).

samruddhikhandale:
> @samruddhikhandale is working on the template https://github.com/devcontainers/templates/pull/39 which can get merged now that the images are released.

.NET Templates have been updated & deployed yesterday.

samruddhikhandale:
Codespaces is now using .NET 7 as default.

Closing as features/templates/images/codespaces have been updated 🚀

PR #201, opened 2022-11-10
Add .NET 7

Chuxel:
Resolves https://github.com/devcontainers/images/issues/200

### DIFF
--- a/src/dotnet/.devcontainer/Dockerfile
+++ b/src/dotnet/.devcontainer/Dockerfile
@@ -1,10 +1,2 @@
-# [Choice] .NET version: 7.0-bullseye, 6.0-bullseye, 3.1-bullseye, 7.0-jammy, 6.0-focal, 3.1-focal
 ARG VARIANT=7.0-bullseye-slim
 FROM mcr.microsoft.com/dotnet/sdk:${VARIANT}
-
-# [Optional] Uncomment this section to install additional OS packages.
-# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
-#     && apt-get -y install --no-install-recommends <your-package-list-here>
-
-# [Optional] Uncomment this line to install global node packages.
-# RUN su vscode -c "source /usr/local/share/nvm/nvm.sh && npm install -g <your-package-here>" 2>&1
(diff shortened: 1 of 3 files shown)


src/dotnet/manifest.json
```diff
@@ -1,18 +1,28 @@
...
 		"rootDistro": "debian",
 		"tags": [
 			"dotnet:${VERSION}-${VARIANT}"
 		],
 		"architectures": {
+			"7.0-bullseye-slim": [
+				"linux/amd64",
+				"linux/arm64"
+			],
+			"7.0-jammy": [
```
samruddhikhandale, line 22: Question - Previously we were publishing `x.0-focal`, are we now publishing `x.0-jammy` as it is latest for ubuntu now ?

src/dotnet/.devcontainer/Dockerfile
```diff
@@ -1,5 +1,5 @@
-# [Choice] .NET version: 6.0-bullseye, 3.1-bullseye, 6.0-focal, 3.1-focal
-ARG VARIANT=6.0-bullseye-slim
+# [Choice] .NET version: 7.0-bullseye, 6.0-bullseye, 3.1-bullseye, 7.0-jammy, 6.0-focal, 3.1-focal
```
samruddhikhandale, line 1: Should we clean such comments in a future PR? I understand they were previously used by the dev containers extension. However, we won't need them now.

samruddhikhandale (commented):
Great! Thank you for the contribution 👏

src/dotnet/manifest.json
Chuxel: .NET 7 only has a jammy version, not focal.  See https://mcr.microsoft.com/en-us/product/dotnet/sdk/tags

The do latest stable Debian and latest stable Ubuntu at time of release.  For 7, that's Jammy and Bullseye.  That's good since Focal has problems on arm64 on macs.

Chuxel: We could, however, add .NET 6 jammy I think.

src/dotnet/.devcontainer/Dockerfile
Chuxel: Done! Removed all comments for that matter.

Chuxel:
Related PR for `universal` is #203

jkeech (approved):

PR #203, opened 2022-11-11
Bump universal .NET versions to 6 and 7

Chuxel:
Related to #200

This is likely at least partly dependent on https://github.com/devcontainers/features/pull/269, but given `installUsingApt` is set to false, it may work without this modification. For consistency, it might be best to wait, however.

### DIFF
--- a/src/universal/.devcontainer/devcontainer.json
+++ b/src/universal/.devcontainer/devcontainer.json
@@ -10,9 +10,9 @@
             "gid": "1000"
         },
         "ghcr.io/devcontainers/features/dotnet:1": {
-            "version": "6",
+            "version": "7",
             "installUsingApt": "false",
-            "additionalVersions": "3"
+            "additionalVersions": "6"
         },
         "ghcr.io/devcontainers/features/hugo:1": {
             "version": "latest"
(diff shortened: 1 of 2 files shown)


src/universal/.devcontainer/devcontainer.json
```diff
@@ -10,9 +10,9 @@
             "gid": "1000"
         },
         "ghcr.io/devcontainers/features/dotnet:1": {
-            "version": "6",
+            "version": "7",
```
samruddhikhandale, line 13: Can we bump the minor version in here ? thanks! 
https://github.com/devcontainers/images/blob/e9f05898cd9fa02cb747bb502cf4e82525791966/src/universal/manifest.json#L2

Chuxel: Done!

Chuxel:
@samruddhikhandale Done!

samruddhikhandale (approved):
Awesome, thank you! 🚀 

image


Looks like the oryx tests for Python are failing which isn't related to this PR changes. I'll fix them in a separate PR. 

https://github.com/devcontainers/images/actions/runs/3464473640/jobs/5786442732#step:3:29798

// cc - @jungaretti

PR #216, opened 2022-11-17
fix version history - Fails to generate info for  `rust` image

samruddhikhandale:
Fixes - https://github.com/devcontainers/images/actions/runs/3464602488/jobs/5789759137#step:5:11948

`rls` is deprecated (Details [here](https://github.com/rust-lang/rls#%EF%B8%8F-rls-is-no-longer-supported)) & hence the `rls --version` command is failing to run. 

Fixing that!

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -376,7 +376,6 @@ For example:
 
 ```json
 "cargo": {
-    "rls": null,
     "rustfmt": null,
     "rust-analysis": "rustc --version",
     "rust-src": "rustc --version",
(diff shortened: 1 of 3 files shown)


edgonmsft (approved):
LGTM

Should something replace this?

samruddhikhandale:
According to [rust-lang/rls](https://github.com/rust-lang/rls#%EF%B8%8F-rls-is-no-longer-supported) repo, `rls` is replaced with `rust-analyzer`. We already have it installed as a VS Code Extension. So we are good 👍

Issue #187, opened 2022-10-31
Sdkman cannot be used with JVM in codespaces


dsyer:
Sdkman is installed by default in the base images, which is potentially quite useful, but it turns out that the base container/VM is doing something special with `${HOME}/.java` which conflicts with it.

In codespaces:

```
$ echo $PATH | tr : '\n' | grep java
/home/codespace/.java/current/bin
/usr/local/sdkman/candidates/java/current/bin
```

In a devcontainer created locally (just with default settings for Java):

```
$ echo $PATH | tr : '\n' | grep java
/usr/local/sdkman/candidates/java/current/bin
/usr/local/sdkman/candidates/java/current/bin
```

The net result is that if you `sdk install java xxx` and `sdk use java xxx` it isn't actually the default java because of the way the `PATH` is set up in the codespace.

samruddhikhandale:
Hi 👋 

I tried to reproduce in a codespace built with `mcr.microsoft.com/devcontainers/universal` image. I was able to successfully set a new default java version.

Attempt 1  - `Do you want java 19.0.1-oracle to be set as default? (Y/n): y`
![image](https://user-images.githubusercontent.com/24955913/199573537-e3a63d0b-7dff-4257-bd27-b4b69030658b.png)


Attempt 2 
![image](https://user-images.githubusercontent.com/24955913/199574484-e712dd9a-5c07-4949-83ad-919138843747.png)


@dsyer could you use the `sdk default java xxx` command instead? It worked for me.

samruddhikhandale:
ℹ️ We are creating a symlink `/home/codespace/.java/current -> /usr/local/sdkman/candidates/java/current`. Hence, whichever java version resides in `/usr/local/sdkman/candidates/java/current` should get reflected with PATH `/home/codespace/.java/current`

dsyer:
You must be using some different configuration than the default (for me):

```
$ ls -l ~/.java/current
lrwxrwxrwx 1 codespace codespace 17 Oct 20 13:07 /home/codespace/.java/current -> /opt/java/11.0.14
$ sdk use java 17.0.5-tem 

Using java version 17.0.5-tem in this shell.
$ ls -l ~/.java/current
lrwxrwxrwx 1 codespace codespace 17 Oct 20 13:07 /home/codespace/.java/current -> /opt/java/11.0.14
$ java -version
openjdk version "11.0.14.1" 2022-02-08 LTS
[...]

PR #220, opened 2022-11-18
Universal: Fixes NVM in a Node project

samruddhikhandale:
Addresses - https://github.com/devcontainers/images/issues/187

For `universal` image, node is first found at `/home/codespace/.nodejs/current` which is sym linked to `usr/local/share/nvm/current`

However, for Node projects, oryx installs another version of node and installs it in `/opt/nodejs` (from source) & links to `/home/codespace/.nodejs/current`. Hence, if the user then tries to set a default node version with nvm, it doesn't work.

This PR makes nvm to work even after oryx installs another version. However, the node version installed by oryx is not set as default. Created another PR on oryx repo to fix this --> https://github.com/microsoft/Oryx/pull/1692

### DIFF
--- a/src/universal/.devcontainer/local-features/setup-user/devcontainer-feature.json
+++ b/src/universal/.devcontainer/local-features/setup-user/devcontainer-feature.json
@@ -4,7 +4,7 @@
     "containerEnv": {
         "RUBY_HOME": "/usr/local/rvm/rubies/default",
         "JAVA_ROOT": "/home/codespace/.java",
-        "NODE_ROOT": "/home/codespace/.nodejs",
+        "NODE_ROOT": "/home/codespace/nvm",
         "PHP_ROOT": "/home/codespace/.php",
         "PYTHON_ROOT": "/home/codespace/.python",
         "RUBY_ROOT": "/home/codespace/.ruby",
@@ -20,7 +20,7 @@
         "RAILS_DEVELOPMENT_HOSTS": ".githubpreview.dev,.app.github.dev",
         "GOROOT": "/usr/local/go",
         "JUPYTERLAB_PATH": "/home/codespace/.local/bin",
-        "PATH": "/home/codespace/.dotnet:/home/codespace/.nodejs/current/bin:/home/codespace/.php/current/bin:/home/codespace/.python/current/bin:/home/codespace/.java/current/bin:/home/codespace/.ruby/current/bin:/home/codespace/.local/bin:${PATH}"
+        "PATH": "/home/codespace/.dotnet:/home/codespace/nvm/current/bin:/home/codespace/.php/current/bin:/home/codespace/.python/current/bin:/home/codespace/.java/current/bin:/home/codespace/.ruby/current/bin:/home/codespace/.local/bin:${PATH}"
     },
     "install": {
         "app": "",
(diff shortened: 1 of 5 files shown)


src/universal/.devcontainer/local-features/setup-user/devcontainer-feature.json
```diff
@@ -4,7 +4,7 @@
     "containerEnv": {
         "RUBY_HOME": "/usr/local/rvm/rubies/default",
         "JAVA_ROOT": "/home/codespace/.java",
-        "NODE_ROOT": "/home/codespace/.nodejs",
+        "NODE_ROOT": "/home/codespace/nvm",
```
jungaretti, line 7: Should this be `.nvm` to match the other language-specific directories?

src/universal/.devcontainer/local-features/setup-user/install.sh
```diff
@@ -30,9 +30,8 @@ ln -snf /usr/local/oryx/* /opt/oryx
 
 # For the universal image, oryx build tool installs the detected platforms in /home/codespace/*. Hence, linking current platforms to the /home/codespace/ path and adding it to the PATH.
 # This ensures that whatever platfornm versions oryx detects and installs are set as root.
-NODE_PATH="/home/codespace/.nodejs/current"
-mkdir -p /home/codespace/.nodejs
-ln -snf /usr/local/share/nvm/current $NODE_PATH
+NODE_PATH="/home/codespace/nvm/current"
+ln -snf /usr/local/share/nvm /home/codespace
```
jungaretti, line 34: Should this link to `~/nvm` or `~/nvm/current`?

Either way, I'd suggest specifying the directory after `/home/codespace` for readability.

src/universal/test-project/sample/node/package.json
```diff
@@ -0,0 +1,4 @@
+{
+    "name": "test-package",
+    "version": "1.0.0"
```
jungaretti, line 3: Should we include a dependency to make sure that Oryx installs them properly?

Here's a simple node project: https://github.com/jungaretti/playground/tree/simple-node

If Oryx is working, then 3 of these commands will succeed:

```bash
cd ./sample/node
oryx build .
npm run test
```

src/universal/.devcontainer/local-features/setup-user/devcontainer-feature.json
samruddhikhandale: I agree that it should be language-specific however, This needs to be `nvm` as we are symlinking `/usr/local/share/nvm --> /home/codespace`

The nvm folder is named `nvm` by default, which we can't change. Hence, in order for oryx and nvm to work together, we need this sym link.

src/universal/.devcontainer/local-features/setup-user/install.sh
samruddhikhandale: We can't do that, the symlink needs to be exactly how it is in this PR. Else, nvm and oryx won't work together 😓 

Currently, we had that, what you are proposing but that's what is breaking sdkman

src/universal/test-project/sample/node/package.json
samruddhikhandale: Only with this PR, the tests would fail until https://github.com/microsoft/Oryx/pull/1692 is merged in.
I was planning to add the tests later.

edgonmsft (approved):

PR #222, opened 2022-11-22
Universal: Remove git installation from source

samruddhikhandale:
Git is already installed from `apt`. Remove, re-installation from git source.

ℹ️  The upstream git has security patch applied!

### DIFF
--- a/src/universal/.devcontainer/Dockerfile
+++ b/src/universal/.devcontainer/Dockerfile
@@ -54,6 +54,7 @@ RUN LANG="C.UTF-8" \
         tk-dev \
         uuid-dev \
         curl \
+        gettext \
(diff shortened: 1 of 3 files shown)


src/universal/.devcontainer/Dockerfile
```diff
@@ -54,6 +54,7 @@ RUN LANG="C.UTF-8" \
         tk-dev \
         uuid-dev \
         curl \
+        gettext \
```
samruddhikhandale, line 57: When git was accidentally switched from apt feeds, there were user threads for missing `gettext`. As this package is expected, installing it here.

joshspicer (approved):

Issue #189, opened 2022-11-01
Universal Image ignores USERNAME variable and enforces codespace as user


rastislav-vcrs:
https://github.com/devcontainers/images/blob/de5005e033eaa4dfda48bdcf0854730c491116cf/src/universal/.devcontainer/local-features/setup-user/install.sh#L33

While reading on this script, all _user_ paths are hardcoded to `/home/codespace`. However, should that be `/home/${USERNAME}` instead to honor `USERNAME=` variable selection at the start of the file?

https://github.com/devcontainers/images/blob/de5005e033eaa4dfda48bdcf0854730c491116cf/src/universal/.devcontainer/local-features/setup-user/install.sh#L7

samruddhikhandale:
Hi 👋 

Thanks for the feedback. I agree that we should change it to use the `USERNAME ` variable 👍 
I would pick this up soon. 👍

PR #223, opened 2022-11-22
Universal: Replace codespace with $USERNAME

samruddhikhandale:
Closes: https://github.com/devcontainers/images/issues/189


joshspicer (approved):
Good catch!

PR #249, opened 2022-12-06
Fix nvm install shell in comments

ShuiRuTian:
Not 100% sure, but there is no `nvm.sh` but `install.sh` in the nvm folder.


Kingparis1 (approved):
It's really nice please keep it up

PR #252, opened 2022-12-08
php: install more extensions, fixes #196

### LINKED ISSUE #196: [PHP] Add more extensions by default
Currently, the docker image for php contains almost no extensions. Maybe we should preinstall popular ones.

We can use also https://github.com/mlocati/docker-php-extension-installer it's better than calling pecl by own 😅

shyim:
- Changed the default variant to PHP 8 (7 is EOL)
- Installed some "default" extensions, with [install-php-extensions](https://github.com/mlocati/docker-php-extension-installer)
- Copy composer from official docker image

### DIFF
--- a/src/php/.devcontainer/Dockerfile
+++ b/src/php/.devcontainer/Dockerfile
@@ -1,19 +1,17 @@
 # [Choice] PHP version (use -bullseye variants on local arm64/Apple Silicon): 8-apache-bullseye, 8.1-apache-bullseye, 8.0-apache-bullseye, 7-apache-bullseye, 7.4-apache-bullseye, 8-apache-buster, 8.1-apache-buster, 8.0-apache-buster, 7-apache-buster, 7.4-apache-buster
-ARG VARIANT=7-apache-bullseye
+ARG VARIANT=8-apache-bullseye
 FROM php:${VARIANT}
 
-# Install xdebug
-RUN yes | pecl install xdebug \
-    && echo "zend_extension=$(find /usr/local/lib/php/extensions/ -name xdebug.so)" > /usr/local/etc/php/conf.d/xdebug.ini \
-    && echo "xdebug.mode = debug" >> /usr/local/etc/php/conf.d/xdebug.ini \
-    && echo "xdebug.start_with_request = yes" >> /usr/local/etc/php/conf.d/xdebug.ini \
-    && echo "xdebug.client_port = 9000" >> /usr/local/etc/php/conf.d/xdebug.ini \
-    && rm -rf /tmp/pear
+COPY --from=mlocati/php-extension-installer /usr/bin/install-php-extensions /usr/local/bin/
+COPY --from=composer/composer:latest-bin /composer /usr/bin/composer
 
-# Install composer
-RUN curl -sSL https://getcomposer.org/installer | php \
-    && chmod +x composer.phar \
-    && mv composer.phar /usr/local/bin/composer
+# Install some php extensions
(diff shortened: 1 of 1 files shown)


shyim:

shyim:
I cannot see the warnings as I cannot access the pipeline. https://github.com/devcontainers/images/pull/252/checks?check_run_id=9977454933

src/php/.devcontainer/Dockerfile
```diff
@@ -1,19 +1,17 @@
...
 FROM php:${VARIANT}
 
-# Install xdebug
-RUN yes | pecl install xdebug \
-    && echo "zend_extension=$(find /usr/local/lib/php/extensions/ -name xdebug.so)" > /usr/local/etc/php/conf.d/xdebug.ini \
-    && echo "xdebug.mode = debug" >> /usr/local/etc/php/conf.d/xdebug.ini \
-    && echo "xdebug.start_with_request = yes" >> /usr/local/etc/php/conf.d/xdebug.ini \
-    && echo "xdebug.client_port = 9000" >> /usr/local/etc/php/conf.d/xdebug.ini \
-    && rm -rf /tmp/pear
+COPY --from=mlocati/php-extension-installer /usr/bin/install-php-extensions /usr/local/bin/
```
samruddhikhandale, line 5: Thanks for opening the PR. Much appreciated! ✨ 

For the core images in this repo, we only include things that are mandatory to the core workflow because anyone can extend them either using a Dockerfile or dev container [features](https://containers.dev/features). Also, you could easily publish your own image with the dev container CLI action.

shyim: What about installing xdebug only as before, but leave this script here.

I would adjust then the README how to install more extensions using that script

src/php/.devcontainer/Dockerfile
```diff
@@ -1,19 +1,17 @@
...
+COPY --from=composer/composer:latest-bin /composer /usr/bin/composer
+
+# Install some php extensions
+RUN IPE_GD_WITHOUTAVIF=1 install-php-extensions bz2 bcmath gd intl mysqli pdo_mysql pdo_pgsql sockets gmp soap zip pcntl redis xsl calendar xdebug
```
samruddhikhandale, line 9: For these utilities, we suggest creating a dev container Feature that could be referenced. Look at https://github.com/devcontainers/feature-starter for getting started. You could create your own Feature in your own repo and share with the community if you'd like.

We are also thinking about composite Features to be able to better link Features together.

samruddhikhandale (commented):
Thanks for opening the PR. Much appreciated! ✨ 

Shared some thoughts in the PR about creating your own Feature or Image to achieve this.

src/php/.devcontainer/Dockerfile
samruddhikhandale: I think creating a "php-extension-installer" Feature would be more beneficial in this scenario as we trying to only keep the mandatory contents in the images.

Else, you could simply use a dockerfile and add these same exact lines to it. 😄

shyim: Is there any guide when to use an image or a feature?

src/php/.devcontainer/Dockerfile
shyim: That script can take up to 5 minutes. I am not sure if this should be really a containers feature

shyim:
I give up improving the default PHP image.

PR #262, opened 2022-12-12
Add `preview.app.github.dev` to `RAILS_DEVELOPMENT_HOSTS`

joshaber:
Caught by @ghiculescu over at https://github.com/microsoft/vscode-dev-containers/pull/1626#issuecomment-1345449615.

I had missed the `preview` prefix when I had updated it before. Leaving `app.github.dev` in place since that will be the final destination of port forwarding.


samruddhikhandale (approved):
Looks good, thanks!

PR #280, opened 2022-12-21
clean out stuff I won't use

ianepreston:


ianepreston:
crap, forgot this was a fork >_<
just messing around with my own copy

PR #284, opened 2022-12-27
Universal: Update 'npm:minimist' due to CVE-2020-7598 & CVE-2021-44906

samruddhikhandale:
More details - https://github.com/advisories/GHSA-vh95-rmgr-6w4m & https://github.com/advisories/GHSA-xvch-5gv4-984h

### DIFF
--- a/src/universal/.devcontainer/local-features/setup-user/install.sh
+++ b/src/universal/.devcontainer/local-features/setup-user/install.sh
@@ -59,7 +59,7 @@ cd /usr/local/share/nvm/versions/node/v14*/lib/node_modules/npm
 npm install ${NPM_PACKAGES_LIST}
 
 # Temporary due to minimist: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-44906 & https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-7598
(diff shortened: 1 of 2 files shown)


src/universal/.devcontainer/local-features/setup-user/install.sh
```diff
@@ -58,6 +58,10 @@ NPM_PACKAGES_LIST="decode-uri-component
 cd /usr/local/share/nvm/versions/node/v14*/lib/node_modules/npm
 npm install ${NPM_PACKAGES_LIST}
 
+# Temporary due to minimist: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-44906 & https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-7598
+cd /usr/local/share/nvm/versions/node/v14.21.2/lib/node_modules/npm/node_modules/tacks
```
chrmarti, line 62: Should the `v14.21.2` path segment be looked up for when it changes or is that the only version that matters and is always pre-installed? (Same below.)

chrmarti (changes requested):

src/universal/.devcontainer/local-features/setup-user/install.sh
samruddhikhandale: Good catch. Looks like i missed adding a `*`. Updated.

chrmarti (approved):
👍

PR #307, opened 2023-01-05
Common-utils Feature: Pin to major version 2

samruddhikhandale:
Ref: https://github.com/devcontainers/features/pull/326

### DIFF
--- a/src/base-debian/.devcontainer/devcontainer.json
+++ b/src/base-debian/.devcontainer/devcontainer.json
@@ -7,8 +7,8 @@
 		"ghcr.io/devcontainers/features/common-utils:2": {
 			"installZsh": "true",
 			"username": "vscode",
-			"uid": "1000",
-			"gid": "1000",
+			"userUid": "1000",
+			"userGid": "1000",
 			"upgradePackages": "true"
 		},
 		"ghcr.io/devcontainers/features/git:1": {
--- a/src/base-debian/.devcontainer/devcontainer.json
+++ b/src/base-debian/.devcontainer/devcontainer.json
@@ -4,7 +4,7 @@
 		"context": "."
 	},
 	"features": {
-		"ghcr.io/devcontainers/features/common-utils:1": {
+		"ghcr.io/devcontainers/features/common-utils:2": {
 			"installZsh": "true",
 			"username": "vscode",
 			"uid": "1000",
--- a/src/base-debian/test-project/test.sh
+++ b/src/base-debian/test-project/test.sh
@@ -19,5 +19,8 @@ cd /tmp && git clone https://github.com/devcontainers/feature-starter.git
 cd feature-starter
 check "perl" bash -c "git -c grep.patternType=perl grep -q 'a.+b'"
 
+check "Oh My Zsh! theme" test -e $HOME/.oh-my-zsh/custom/themes/devcontainers.zsh-theme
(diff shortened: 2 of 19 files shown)


eitsupi (commented):
I believe the `uid` and `gid` option names have changed, so must be update here. (same #308)
https://github.com/devcontainers/features/blob/a9a7a25d60585ce2145374ec6825cbdc7f16cb90/src/common-utils/devcontainer-feature.json#L40-L57

I am also wondering if it is possible to change the `meta.env`'s path.
https://github.com/devcontainers/features/blob/a9a7a25d60585ce2145374ec6825cbdc7f16cb90/src/common-utils/bin/devcontainer-info#L7

samruddhikhandale:
> I believe the uid and gid option names have changed, so must be update here. (same https://github.com/devcontainers/images/pull/308)
https://github.com/devcontainers/features/blob/a9a7a25d60585ce2145374ec6825cbdc7f16cb90/src/common-utils/devcontainer-feature.json#L40-L57

Yep, thanks for catching. Updated

samruddhikhandale:
> I am also wondering if it is possible to change the meta.env's path.
https://github.com/devcontainers/features/blob/a9a7a25d60585ce2145374ec6825cbdc7f16cb90/src/common-utils/bin/devcontainer-info#L7

I think we could update that, it would mostly need an update to the vscdc tool [here](https://github.com/devcontainers/images/blob/main/build/src/prep.js#L56). @Chuxel any concerns?

Chuxel:
> > I am also wondering if it is possible to change the meta.env's path.
> > https://github.com/devcontainers/features/blob/a9a7a25d60585ce2145374ec6825cbdc7f16cb90/src/common-utils/bin/devcontainer-info#L7
> 
> I think we could update that, it would mostly need an update to the vscdc tool [here](https://github.com/devcontainers/images/blob/main/build/src/prep.js#L56). @Chuxel any concerns?

Yep. Looks like @eitsupi did a PR already to update the common-utils `devcontainer-info` command to check the `dev-containers` folder in addition to `vscode-dev-containers`, so updating the image generation should be all that is required!

samruddhikhandale:
> > > I am also wondering if it is possible to change the meta.env's path.
> > > https://github.com/devcontainers/features/blob/a9a7a25d60585ce2145374ec6825cbdc7f16cb90/src/common-utils/bin/devcontainer-info#L7
> > 
> > 
> > I think we could update that, it would mostly need an update to the vscdc tool [here](https://github.com/devcontainers/images/blob/main/build/src/prep.js#L56). @Chuxel any concerns?
> 
> Yep. Looks like @eitsupi did a PR already to update the common-utils `devcontainer-info` command to check the `dev-containers` folder in addition to `vscode-dev-containers`, so updating the image generation should be all that is required!

Great! I'll create another PR once this one's merged, thanks!

edgonmsft (dismissed):

src/base-debian/.devcontainer/devcontainer.json
```diff
@@ -4,11 +4,11 @@
 		"context": "."
 	},
 	"features": {
-		"ghcr.io/devcontainers/features/common-utils:1": {
+		"ghcr.io/devcontainers/features/common-utils:2": {
```
samruddhikhandale, line 7: Blocked by https://github.com/devcontainers/features/pull/388

PR #308, opened 2023-01-05
Alpine: Use common-utils Feature instead of a lib script

samruddhikhandale:
Ref: https://github.com/devcontainers/features/pull/326


src/base-alpine/.devcontainer/devcontainer.json
```diff
@@ -3,15 +3,23 @@
...
 	// "forwardPorts": [],
 
 	// Use 'postCreateCommand' to run commands after the container is created.
 	// "postCreateCommand": "uname -a",
 
 	// Replace when using a ptrace-based debugger like C++, Go, and Rust
-	// "runArgs": [ "--init", "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined" ],
-
-	// Set `remoteUser` to `root` to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
-	"remoteUser": "vscode"
```
Chuxel, line 16: I assume you still need `"remoteUser": "vscode"`, correct?

samruddhikhandale: Oops, my bad

Chuxel (dismissed):

src/base-alpine/.devcontainer/devcontainer.json
```diff
@@ -3,6 +3,17 @@
 		"dockerfile": "./Dockerfile",
 		"context": "."
 	},
+
+	"features": {
+		"ghcr.io/devcontainers/features/common-utils:2": {
+			"installZsh": "true",
+			"username": "vscode",
+			"uid": "1000",
+			"gid": "1000",
```
eitsupi, line 12:
```suggestion
			"userUid": "1000",
			"userGid": "1000",
```

Chuxel: Good catch! 1000 is the default so it would work, but you're right we should fix that.

samruddhikhandale: Ah, I missed that. Thanks!

edgonmsft (approved):

src/base-alpine/.devcontainer/devcontainer.json
```diff
@@ -3,6 +3,17 @@
 		"dockerfile": "./Dockerfile",
 		"context": "."
 	},
+
+	"features": {
```
samruddhikhandale, line 7: Waiting for https://github.com/devcontainers/features/pull/388

PR #312, opened 2023-01-06
Added codespaces jupyterlab extension to list of machine learning packages to install in universal image

8Gitbrix:
The purpose of this change is to add the [codespaces jupyterlab extension](https://github.com/github/codespaces-jupyterlab) package to the universal image. Since the image is enabled by default for codespaces, this package will already be preinstalled for codespaces users, and save time for users opening a codespace in a jupyterlab editor.

### DIFF
--- a/cgmanifest.json
+++ b/cgmanifest.json
@@ -1216,6 +1216,15 @@
         },
         {
             "Component": {
+                "Type": "Pip",
+                "Pip": {
+                    "Name": "codespaces_jupyterlab",
+                    "Version": "0.2.1"
+                }
+            }
+        },
+        {
+            "Component": {
                 "Type": "RubyGems",
                 "RubyGems": {
                     "Name": "rake",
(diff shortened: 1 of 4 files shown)


cgmanifest.json
```diff
@@ -1214,6 +1214,15 @@
                 }
             }
         },
+        {
```
samruddhikhandale, line 1217: This file is auto-generated from images' [manifest.json](https://github.com/devcontainers/images/blob/main/src/universal/manifest.json). Can we revert this change and instead add it in here 👇 ? thanks

https://github.com/devcontainers/images/blob/73b30a9f2fbbdaf54fb6fdda7cd7dce86ca63f5f/src/universal/manifest.json#L110-L122

samruddhikhandale (changes requested):
Thanks looks good, left one minor comment.

Issue #385, opened 2023-01-29
mcr.microsoft.com/devcontainers/jekyll:0.1.18 doesn't work when Gemfile.lock is present


tudortimi:
I have a Jekyll project with a `Gemfile.lock`. If I try to use version 0.1.18 of the image, the container fails during the post create command.

Here's a `devcontainer,json`

```
{
	"name": "Jekyll",
	"image": "mcr.microsoft.com/devcontainers/jekyll:0.1.18-2.7-bullseye"
}
```
and a `Gemfile.lock`

```
GEM
  remote: https://rubygems.org/
  specs:
    addressable (2.8.0)
      public_suffix (>= 2.0.2, < 5.0)
    colorator (1.1.0)
    concurrent-ruby (1.1.9)
    em-websocket (0.5.2)
      eventmachine (>= 0.12.9)
      http_parser.rb (~> 0.6.0)
    eventmachine (1.2.7)
    faraday (1.8.0)
      faraday-em_http (~> 1.0)
      faraday-em_synchrony (~> 1.0)
      faraday-excon (~> 1.1)
      faraday-httpclient (~> 1.0.1)
      faraday-net_http (~> 1.0)
      faraday-net_http_persistent (~> 1.1)
      faraday-patron (~> 1.0)
      faraday-rack (~> 1.0)
      multipart-post (>= 1.2, < 3)
      ruby2_keywords (>= 0.0.4)
    faraday-em_http (1.0.0)
    faraday-em_synchrony (1.0.0)
    faraday-excon (1.1.0)
    faraday-httpclient (1.0.1)
    faraday-net_http (1.0.1)
    faraday-net_http_persistent (1.2.0)
    faraday-patron (1.0.0)
    faraday-rack (1.0.0)
    ffi (1.15.4)
    forwardable-extended (2.6.0)
    http_parser.rb (0.6.0)
    i18n (1.8.11)
      concurrent-ruby (~> 1.0)
    jekyll (4.2.1)
      addressable (~> 2.4)
      colorator (~> 1.0)
      em-websocket (~> 0.5)
      i18n (~> 1.0)
      jekyll-sass-converter (~> 2.0)
      jekyll-watch (~> 2.0)
      kramdown (~> 2.3)
      kramdown-parser-gfm (~> 1.0)
      liquid (~> 4.0)
      mercenary (~> 0.4.0)
      pathutil (~> 0.9)
      rouge (~> 3.0)
      safe_yaml (~> 1.0)
[...]

PR #390, opened 2023-01-31
Jekyll - Fix bug "cannot open ./.devcontainer/post-create.sh: No such file"

samruddhikhandale:
Reported in https://github.com/devcontainers/images/issues/385#issuecomment-1409517940


joshspicer (approved):
nice!

PR #391, opened 2023-01-31
Retry az login to fix dev release failures

samruddhikhandale:
Az login has a expiry of 3 hours, on of the job for dev release is taking longer than that. Hence, it fails to push the image.
This attempts to fix it! 

https://github.com/devcontainers/images/actions/runs/4044601960


joshspicer (approved):
Probably not a big deal, but we could check first and only attempt to re-login if necessary.  This is likely fast enough it doesn't matter though.

samruddhikhandale:
It takes couple of seconds, so should be fine. I'll get back to improving the logic if I find it degrading release perf.

Issue #407, opened 2023-02-06
Make UTF-8 locale the default


rhuitl:
I'm investigating the root cause for https://github.com/microsoft/vscode-java-dependency/issues/712.

It looks like we might have to set the environment variable `LANG=C.UTF-8` so that all processes running in the container have UTF-8 support. It appears that e.g. Java will only be able to use UTF-8 if the locale is set.

We'd still have to verify if this is indeed fixing the problem, but I wanted to get your opinion first.

That variable is already set for one command at build time (https://github.com/devcontainers/images/blob/main/src/universal/.devcontainer/Dockerfile#L15) but we'd have to turn it into a `ENV` line so it also takes effect when the container is run.

rhuitl:
This change fixes the issue for me: https://github.com/rhuitl/vscode-java-bug/commit/196bbdeb3f07ac2bbc7e19a468eeb6c9525a4196

samruddhikhandale:
Hi @rhuitl 👋 

This makes sense to me. Opened https://github.com/devcontainers/images/pull/417

bamurtaugh:
Thanks again for filing!

Can we close this one now that @samruddhikhandale's PR has been merged?

rhuitl:
Yes, I think so! Thank you :)

samruddhikhandale:
The PR's merged, however, the changes are not reflected in the `universal` image. We need an images release (planning to do one today/tomorrow), hence, I had intentionally kept this PR open for tracking purpose 😅

samruddhikhandale:
I'll close this PR once the images are updated with the changes, hope that's fine. Thanks!

samruddhikhandale:
Closing as the `universal` image is updated 🚀

PR #417, opened 2023-02-09
Universal: Set LANG="C.UTF-8" as ENV

samruddhikhandale:
Ref: https://github.com/devcontainers/images/issues/407


rhuitl:
Cool! Ideally all devcontainers would support UTF-8. I think some have LANG set, others LANG and LC_ALL, and some not at all and maybe have it from their parent image (e.g. node?) or maybe not...

joshspicer (approved):

PR #418, opened 2023-02-09
Drop bandit, pycodestyle, pydocstyle, and yapf from the Python image

brettcannon:
All of those tools are being dropped from the Python extension for VS Code due to lack of usage.


samruddhikhandale:
Hi 👋 

Thanks for the information. To remove these utils from the python image, they first need to be removed from the Python Feature [here](https://github.com/devcontainers/features/blob/main/src/python/install.sh#L29)

ℹ️ `dev.md` & `manifest.json` are autogenerated docs which will get overwritten unless we remove them from the Python Feature. The only change needed for images will be removing the checks from `test.sh` and releasing new images. 😄 

// cc @Chuxel / @bamurtaugh / @craiglpeters Any thoughts/concerns for removing these utils from the Python Feature (which will end up getting removed from the images as well) ? 

I could help with the cleanup if there are no concerns.

bamurtaugh:
I'm all in favor of cleanup wherever possible like this!

@brettcannon How are you notifying users of this change in the Python extension? Since it sounds like this could be potentially breaking for a small number of users, curious on the approach. 

As a sanity check we could perform before removing, we could post this PR in the dev container community Slack channel as a heads up and see if anyone from that group would be impacted or knows of a scenario that would be.

brettcannon:
> To remove these utils from the python image, they first need to be removed from the Python Feature [here](https://github.com/devcontainers/features/blob/main/src/python/install.sh#L29)

Already done in https://github.com/devcontainers/features/pull/449 .



> How are you notifying users of this change in the Python extension?

We're not there yet, this is just pre-emptive so users don't start using these tools and then have them disappear on them in a few months.


> As a sanity check we could perform before removing, we could post this PR in the dev container community Slack channel as a heads up and see if anyone from that group would be impacted or knows of a scenario that would be.

Do note this is somewhat of a part 1 of 2. Nearly all the tools left behind will eventually be in separate extensions which ship them in-box. As I said, I'm doing this now so people don't start to rely on things I know are on there way out the door, but I didn't want to be _too_ disruptive by yanking stuff that will work _somehow_, just perhaps via a separate extension.

brettcannon:
For me, the key thing is certain things shouldn't get lit up for the Python extension for VS Code (e.g. don't turn on Pylint on by default, don't direct people towards yapf). Knowing that, how would you prefer me to handle that? Do you just want me to turn off everything or ignore the tools in https://github.com/devcontainers/features/pull/449 instead? I'm just not clear on how I should view this image since it seems to install things based on the Python extension, but if that was just a way to start this and I should view this as independent of the extension that's obviously fine.

bamurtaugh:
> For me, the key thing is certain things shouldn't get lit up for the Python extension for VS Code

This is helpful context. Since these images are for us in any dev container spec-supporting tools, including but not limited to VS Code, I could see the potential for some deviation.

From my understanding, the [change in the Features PR](https://github.com/devcontainers/features/pull/449/files#diff-ff59b43567a04a2385291ff38f89705a0ff16a7bfb5e1e503732f43935099208R63-R71) that updates VS Code settings would make sense to merge in, as this is a VS Code-specific aspect of the Feature setup. But, I think we'd want to leave general support for these libraries in other parts of the image and Feature if they're going to come back in through separate extensions as you mention?

brettcannon:
I'll focus on updating the settings over in https://github.com/devcontainers/features/pull/449/ then and close this out to leave the base image alone.



> I think we'd want to leave general support for these libraries in other parts of the image and Feature if they're still potentially useful libraries (as they're likely going to come back in through separate extensions as you mention)?

Actually, none of the tools I slated for removal in this PR are coming back through separate extensions; they are going to be flat-out dropped from support. The tools I left in will be coming into VS Code via separate extensions.

PR #421, opened 2023-02-13
Universal: Set moby:false as it is no longer supported for unbuntu:focal

samruddhikhandale:
Fixes build failure - https://github.com/devcontainers/images/actions/runs/4164214688/jobs/7211607878#step:5:36028

### DIFF
--- a/src/universal/.devcontainer/devcontainer.json
+++ b/src/universal/.devcontainer/devcontainer.json
@@ -57,7 +57,8 @@
             "version": "latest"
         },
         "ghcr.io/devcontainers/features/docker-in-docker:2": {
-            "version": "latest"
+            "version": "latest",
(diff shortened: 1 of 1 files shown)


src/universal/.devcontainer/devcontainer.json
```diff
@@ -57,7 +57,8 @@
             "version": "latest"
         },
         "ghcr.io/devcontainers/features/docker-in-docker:2": {
-            "version": "latest"
+            "version": "latest",
+            "moby": false
```
samruddhikhandale, line 61: https://github.com/devcontainers/images/actions/runs/4167757752/jobs/7213736394 A recent build was successful, looks like a temporary glitch and probably this isn't needed anymore?

PR #422, opened 2023-02-13
Universal: Removes gradle 7 patches due to newer version

samruddhikhandale:
Fixes - https://github.com/devcontainers/images/actions/runs/4167020613/jobs/7212085138#step:3:40923

### DIFF
--- a/src/universal/.devcontainer/local-features/setup-user/install.sh
+++ b/src/universal/.devcontainer/local-features/setup-user/install.sh
@@ -20,36 +20,6 @@ chmod +x /etc/profile.d/00-restore-env.sh
 
 export DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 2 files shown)


src/universal/.devcontainer/local-features/setup-user/install.sh
```diff
@@ -20,36 +20,6 @@ chmod +x /etc/profile.d/00-restore-env.sh
 
 export DEBIAN_FRONTEND=noninteractive
 
-# Temporary: Replace the current gradle plugins with patched versions
```
samruddhikhandale, line 23: Gradle 8 is available with the patches!

joshspicer (approved):

5168745119118245 (commented):
Jo

PR #429, opened 2023-02-15
History: Update manifest.json to include root 'git' version

samruddhikhandale:
These changes will add a new field to the history files which prints the git version. Similar to https://github.com/devcontainers/images/pull/428/files#diff-dd1e63701c65447d6ad769929357102900436e7c30b76d59c9612c26ba9a6f7aR104


joshspicer (dismissed):
Why do we need `null` here instead of just omitting the `git` property?

samruddhikhandale:
> Why do we need `null` here instead of just omitting the `git` property?

Good point, updated to `"git": {}`

joshspicer (approved):

PR #475, opened 2023-03-06
Fix bad indent

NickBorgers:
Looks like this repo could use some linters as Actions and/or githooks.

I'm looking at possibly pushing for adoption of this spec so willing to do some work. Has linting on the Action pipeline already been implemented? Any specific approach that seems like it would be preferred/better?

### DIFF
--- a/src/python/.devcontainer/devcontainer.json
+++ b/src/python/.devcontainer/devcontainer.json
@@ -13,7 +13,7 @@
         },
         "ghcr.io/devcontainers/features/python:1": "none",
         "ghcr.io/devcontainers/features/node:1": "none",
-		"ghcr.io/devcontainers/features/git:1": {
+        "ghcr.io/devcontainers/features/git:1": {
             "version": "latest",
             "ppa": "false"
         }


eitsupi:
The previous repository, https://github.com/microsoft/vscode-dev-containers, used editorconfig, but it does not appear to be in this repository.
It is possible that some decision has been made.

NickBorgers:
> The previous repository, https://github.com/microsoft/vscode-dev-containers, used editorconfig, but it does not appear to be in this repository. It is possible that some decision has been made.

Looking at the announcement in that README also reveals to me that Devcontainer being an open spec is new. I was wondering how I missed that before...

samruddhikhandale:
> The previous repository, https://github.com/microsoft/vscode-dev-containers, used editorconfig, but it does not appear to be in this repository. It is possible that some decision has been made.

Thanks! Looks like it was missed while migrating images to this repository. I'll open up a PR 👍

samruddhikhandale (approved):
Thanks!

PR #476, opened 2023-03-06
Add .editorconfig

samruddhikhandale:
Addresses https://github.com/devcontainers/images/pull/475#issuecomment-1456149887
From https://github.com/microsoft/vscode-dev-containers/blob/main/.editorconfig

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,23 +1,17 @@
 {
     "image": "mcr.microsoft.com/devcontainers/javascript-node:0-18",
-    "customizations": {
-        "vscode": {
-            "extensions": [
-                "mads-hartmann.bash-ide-vscode",
-                "dbaeumer.vscode-eslint"
-            ]
-        },
-        "codespaces": {
-            "repositories": {
-                "devcontainers/features": {
-                    "permissions": "write-all"
-                }
-            }
-        }
-    },
     "features": {
         "ghcr.io/devcontainers/features/docker-in-docker:2": {},
         "ghcr.io/devcontainers/features/azure-cli:1": {}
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,10 @@
+[*.json]
```
eitsupi, line 1: In general, it seems that `root = true` is added to the project root file.
https://github.com/editorconfig/editorconfig/blob/b52217517b2c5b2c3b6a0daa7260983d409e7e6b/.editorconfig#LL2-L2C12

```suggestion
root = true

[*.json]
```

eitsupi (commented):
It would be useful to add the editorconfig extension for VS Code to the devcontainer.json.
https://github.com/microsoft/vscode-dev-containers/blob/5a084a93b0736ea86395ac99019a5b72a00b6341/.vscode/extensions.json#L6-L9

.devcontainer/devcontainer.json
```diff
@@ -1,23 +1,17 @@
...
                 "mads-hartmann.bash-ide-vscode",
-                "dbaeumer.vscode-eslint"
+                "dbaeumer.vscode-eslint",
+                "EditorConfig.EditorConfig"
             ]
-        },
-        "codespaces": {
-            "repositories": {
-                "devcontainers/features": {
-                    "permissions": "write-all"
```
samruddhikhandale, line 13: I don't believe we need these permissions anymore. @joshspicer is that correct?

joshspicer: Not necessary.  It may be handy at times but definitely ok to remove.

joshspicer (approved):

PR #481, opened 2023-03-08
Update VS Code setting for the Python extension

brettcannon:
Remove tool paths for things that are planned for removal from the Python extension (hopefully) this year. Also explicitly turn linting off by default as they are moving to their own extensions (which already exist), and to minimize noise (e.g. Pylint complaining about imports).

This brings things in line with https://github.com/devcontainers/features/tree/main/src/python .


samruddhikhandale (approved):
Thanks!

PR #516, opened 2023-04-04
Jekyll - EOL ruby 2.7 ; use ruby 3.2 instead

samruddhikhandale:
Ref: https://www.ruby-lang.org/en/downloads/branches/ & https://github.com/devcontainers/images/pull/514


.github/actions/smoke-test/test.sh
```diff
@@ -13,4 +13,4 @@ echo "(*) Docker image details..."
 docker images
 
 # Clean up
-docker rm -f $(docker container ls -f "label=${id_label}" -q)
+# docker rm -f $(docker container ls -f "label=${id_label}" -q)
```
joshspicer, line 16: What's the reason for this?

joshspicer (dismissed):

.github/actions/smoke-test/test.sh
samruddhikhandale: Oops! That was my local debugging, removing

joshspicer (approved):

Issue #443, opened 2023-02-22
Tracking: Alpine 3.14 EOL scheduled for 1st May 2023


samruddhikhandale:
- [ ] Open a PR and deprecate/EOL alpine 3.14
- [x] Update https://github.com/devcontainers/images/issues/90
- [ ] Update templates if necessary

samruddhikhandale:
Closing as completed.

PR #562, opened 2023-05-10
EOL - Alpine 3.14

samruddhikhandale:
Ref: https://github.com/devcontainers/images/issues/443


chrmarti (approved):
LGTM. I guess you will update the manifest version at a later point?

samruddhikhandale:
> I guess you will update the manifest version at a later point?

Yes, for simpler tracking we are now creating a separate PR for version bumps.

PR #569, opened 2023-05-18
[anaconda] Reorg Dockerfile instructions to resolve issue with group permissions and `PIP` 

alexander-smolyakov:
**Devcontainer name**:

- anaconda

**Description**:

This PR addresses the `[Errno 13] Permission denied` error that recently appeared during `anaconda` devcontainer smoke testing. The error relates to the fact that when PIP installs updates for Python packages, it's not preserving group permissions which results in flipping the group write permission bite:

_Without running `pip install`_:

```console
vscode ➜ ~ $ tree -ugp -L 1 /opt/conda/lib/python3.10/site-packages/
/opt/conda/lib/python3.10/site-packages/
...
├── [drwxrwsr-x root     conda   ]  cookiecutter
├── [drwxrwsr-x root     conda   ]  cookiecutter-1.7.3.dist-info
├── [drwxrwsr-x root     conda   ]  cryptography
├── [drwxrwsr-x root     conda   ]  cryptography-39.0.1.dist-info
```

_With running `pip install`_:

```console
vscode ➜ ~ $ tree -ugp -L 1 /opt/conda/lib/python3.10/site-packages/
/opt/conda/lib/python3.10/site-packages/
...
├── [drwxr-sr-x root     conda   ]  cookiecutter
├── [drwxr-sr-x root     conda   ]  cookiecutter-2.1.1.dist-info
├── [drwxr-sr-x root     conda   ]  cryptography
├── [drwxr-sr-x root     conda   ]  cryptography-40.0.2.dist-info
```

**Fix**:

We need to reorg several instructions in `Dockerfile` to ensure the `site-packages` directory has the correct permissions for all packages.

**Changelog**:

- Updated Dockerfile instructions and they order to address issues with PIP and group permission;
- Removed temp security patches that are no longer actual;
- Added missing tests for Python packages version check;

PR #571, opened 2023-05-19
[miniconda] Reorg Dockerfile instructions to resolve issue with group permissions and PIP

alexander-smolyakov:
**Devcontainer name**:

- miniconda

**Description**:

This PR applies the same changes as in #569 but for the `miniconda` devcontainer.

**Changelog**:

- Updated Dockerfile instructions and they order to address issues with PIP and group permission;
- Removed temp security patches that are no longer actual;


samruddhikhandale (approved):
Nice, thanks! 🎉

Issue #486, opened 2023-03-09
Anaconda image has incompatible cryptography and openssl modules, breaking conda package manager


broander:
Hi,

Currently conda is broken in the latest image available (mcr.microsoft.com/devcontainers/anaconda:0-3).  This issue has been going on for a couple of months, but reporting it now that it's blocking installing new dependencies for a project.

If after creating a fresh container one attempts to install anything with conda, including just updating conda with 'conda update conda', the following error occurs: "AttributeError: module 'lib' has no attribute 'OpenSSL_add_all_algorithms'"

There is discussion of the issue here: https://github.com/pyca/cryptography/issues/7959  The root cause is due to incompatible versions of cryptography and openssl via pyopenssl coexisting at once in the same environment.

While one can work around the issue by upgrading pyopenssl using pip, this breaks the conda package manager's ability to successfully complete transactions going forward to upgrade or install new packages, transactions either taking very, very long to complete or failing to verify and ending without executing.

Would you please fix this incompatibility in the latest image and going forward so we can use conda to install packages?   
 A potential solution would be reving the version of anaconda being installed on the source image to one where this incompatibility isn't present from inception.  Thanks!

broander:
Also, I just tried creating a container using the dev image (mcr.microsoft.com/vscode/devcontainers/anaconda:dev-3) and the same issue outlined above is present.

broander:
I tried out the miniconda image this morning (mcr.microsoft.com/devcontainers/miniconda:0-3), and it does not have the same issue, and one is able to install new packages using conda.  However, you might find it interesting that if you try 'conda update conda' with the miniconda package, it does raise an error related to the cryptography package, although a different one than the Anaconda image:

![image](https://user-images.githubusercontent.com/66220841/224071453-a469990c-6450-477b-a561-1836a7e9237d.png)

broander:
@samruddhikhandale Thanks for taking a look at this.  I see you're working through developing your pull request, appreciate it.  I tried your latest solution for the Anaconda image, which is having pip install pyOpenssl.  That does indeed allow 'conda update conda' to succeed.  However, you should add another test to install a package after that, for example 'conda install -c conda-forge pylint'.  That is failing; I also tested with several other common packages on conda-forge.

I also tried your miniconda solution, also fails when you try the same conda-forge command mentioned above and several other common packages as well.

broander:
[...]

PR #580, opened 2023-05-30
Anaconda: Adds back 'anaconda' pkg

samruddhikhandale:
Fixes image history generation - https://github.com/devcontainers/images/actions/runs/5123831476/jobs/9219811428#step:5:263

Since the update to anaconda distribution `2023.03-1`, `continuumio/anaconda3` image no more contains the `anaconda` pkg. The dev container's anaconda image expects that to exist, hence install manually.

### DIFF
--- a/src/anaconda/.devcontainer/Dockerfile
+++ b/src/anaconda/.devcontainer/Dockerfile
@@ -43,6 +43,9 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && echo "conda activate base" >> ~/.bashrc \
     && apt-get clean -y && rm -rf /var/lib/apt/lists/* /tmp/library-scripts/add-notice.sh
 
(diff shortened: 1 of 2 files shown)


src/anaconda/.devcontainer/Dockerfile
```diff
@@ -54,8 +57,6 @@ RUN python3 -m pip install \
     mistune \
     # https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-34141
     numpy \
-    # https://github.com/devcontainers/images/issues/486
-    pyOpenssl \
```
samruddhikhandale, line 58: We have removed patch to the `cryptography` package, hence, we don't need this update.

Ref: https://github.com/devcontainers/images/issues/486#issuecomment-1462609303

// cc @alexander-smolyakov

joshspicer (dismissed):

Issue #653, opened 2023-07-10
Where are the Dockerfiles? 😄


dpprdan:
The https://github.com/devcontainers/images/blob/main/src/python/README.md reads:

> Alternatively, you can use the contents of `Dockerfile` to fully customize the your container's contents or build for a container architecture the image does not support.

I am assuming that this refers to a Dockerfile in this repo? However I cannot find it anywhere (apart from https://github.com/devcontainers/images/blob/main/src/python/.devcontainer/Dockerfile which doesn't seem to be complete? E.g. I don't see where [the dependencies](https://github.com/devcontainers/images/blob/99acaadbc99a6211b3e5452203ffc90b9e63ae75/src/python/manifest.json#L95) are added). 

Put differently: Is it possible to view (example-)Dockerfiles of the final images?

samruddhikhandale:
Hi 👋 

> I am assuming that this refers to a Dockerfile in this repo? 

Yes, that statement is pointing to the [Dockerfile](https://github.com/devcontainers/images/blob/main/src/python/.devcontainer/Dockerfile) you mentioned. 

> However I cannot find it anywhere (apart from https://github.com/devcontainers/images/blob/main/src/python/.devcontainer/Dockerfile which doesn't seem to be complete?

Dev container Features are doing rest of the work, see [here](https://github.com/devcontainers/images/blob/main/src/python/.devcontainer/devcontainer.json#L6-L20). Look at https://containers.dev/implementors/features/ for more information.

> E.g. I don't see where [the dependencies](https://github.com/devcontainers/images/blob/99acaadbc99a6211b3e5452203ffc90b9e63ae75/src/python/manifest.json#L95) are added).

They are installed by the Python Feature, in [here](https://github.com/devcontainers/features/blob/main/src/python/install.sh#L29).

@dpprdan Let me know if that helps clear your doubts, or happy to answer any other questions.

> Alternatively, you can use the contents of Dockerfile to fully customize the your container's contents or build for a container architecture the image does not support.

@Chuxel We copied over the Readme file from https://github.com/microsoft/vscode-dev-containers/tree/main/containers/python-3 when Features didn't exist and `base.dockerfile` was doing all the work. I wonder if we should replace current statement to 👇 (to avoid confusions) or remove that line? Any thoughts? // cc @bamurtaugh 

> Alternatively, you can use the contents of <s> `Dockerfile`</s> `devcontainer` to fully customize the your container's contents or build for a container architecture the image does not support.

[...]

PR #654, opened 2023-07-10
Update docs - Point to `.devcontainer` folder instead of the `Dockerfile` file

samruddhikhandale:
Ref: https://github.com/devcontainers/images/issues/653#issuecomment-1629823470

### DIFF
--- a/src/anaconda/README.md
+++ b/src/anaconda/README.md
@@ -34,7 +34,7 @@ You can decide how often you want updates by referencing a [semantic version](ht
 
 See [history](history) for information on the contents of each version and [here for a complete list of available tags](https://mcr.microsoft.com/v2/devcontainers/anaconda/tags/list).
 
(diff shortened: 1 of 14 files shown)


joshspicer (changes requested):
Are these supposed to be links?

samruddhikhandale:
> Are these supposed to be links?

@joshspicer Yes, From this issue https://github.com/devcontainers/images/issues/653, looks there was a confusion as in which Dockerfile we were referencing to? Hence, linking should avoid any unwanted confusions.

Also, https://github.com/devcontainers/images/issues/653#issuecomment-1629789928

> I kept the last sentence more general to avoid needing to link every readme uniquely, though if we're willing to do that and don't think the links will break, that'd also probably help.

joshspicer (approved):

PR #656, opened 2023-07-12
Docs: polish README.md

dpprdan:
These changes make the README a bit more comprehensible (IMHO) and fixes a few typos.

I propose to discuss the changes for this single README first and then roll the changes out to the others as well (either with this or another PR).

cf #653


src/python/README.md
```diff
@@ -37,13 +37,14 @@ You can decide how often you want updates by referencing a [semantic version](ht
...
 
 See [history](history) for information on the contents of each version and [here for a complete list of available tags](https://mcr.microsoft.com/v2/devcontainers/python/tags/list).
 
-Alternatively, you can use the contents of `Dockerfile` to fully customize the your container's contents or build for a container architecture the image does not support.
+Alternatively, you can use the contents of [.devcontainer](.devcontainer) to fully customize the your container's contents or build for a container architecture the image does not support.
 
-Beyond Python and `git`, this image / `Dockerfile` includes a number of Python tools, `zsh`, [Oh My Zsh!](https://ohmyz.sh/), a non-root `vscode` user with `sudo` access, and a set of common dependencies for development.
+Beyond Python and `git`, this image includes a number of Python tools, `zsh`, [Oh My Zsh!](https://ohmyz.sh/), a non-root `vscode` user with `sudo` access, and a set of common dependencies for development, see [.devcontainer/devcontainer.json](.devcontainer/devcontainer.json).
```
dpprdan, line 47: I added this link because I think it might be helpful to see where these additions to the image happen (if you are, like me currently, new to devcontainers).

dpprdan:
[This sentence](https://github.com/dpprdan/images/blob/5a347becba71f72f0941e8e13ac0cd841874f6bd/src/python/README.md?plain=1#L34C1-L34C127) is still a bit too terse IMHO:

> You can decide how often you want updates by referencing a [semantic version](https://semver.org/) of each image.

Which updates are meant here?

src/python/README.md
```diff
@@ -21,7 +21,7 @@ See **[history](history)** for information on the contents of published images.
 
 ### Configuration
 
-You can directly reference pre-built versions of `Dockerfile` by using the `image` property in `.devcontainer/devcontainer.json` or updating the `FROM` statement in your own `Dockerfile` with one of the following:
+You can directly reference pre-built versions of this image by using the `image` property in `.devcontainer/devcontainer.json` or updating the `FROM` statement in your own `Dockerfile` with one of the following:
```
samruddhikhandale, line 24:
```suggestion
You can directly reference pre-built versions of this [.devcontainer](.devcontainer) by using the `image` property in `.devcontainer/devcontainer.json` or updating the `FROM` statement in your own `Dockerfile` with one of the following:
```

Curious, does this make more sense? 
// cc @bamurtaugh

samruddhikhandale (commented):
Thank you for taking the time to polish the README, we appreciate it ✨ 

Left one comment, besides that looks good to me!

src/python/README.md
bamurtaugh: Thanks for tagging me! I could see either making sense. I think `image` makes a bit more sense to me as we don't really use `.devcontainer` as a term elsewhere.

src/python/README.md
bamurtaugh: Thanks for sharing this! I'm not sure I'd add the devcontainer.json pointer. It's definitely a key file, but not the only one that includes tools or dependencies, so it may be a bit confusing.

bamurtaugh (approved):
Thanks so much for taking the time to put this together, @dpprdan!

bamurtaugh (commented):
Thanks so much for taking the time to put this together, @dpprdan! (updating to "comment" as that's the feedback level I meant initially 😄)

src/python/README.md
dpprdan: I'd definitely stick with `image` here, because that is what is meant in this sentence (IIUC), i.e. the [Docker image](https://docs.docker.com/glossary/#image), not the instructions of how to build that image, i.e. "traditionally" the `Dockerfile`, here the `Dockerfile` + `devcontainer.json`.

`pre-built` is a bit awkward maybe, because images are by definition built, so maybe 

```suggestion
You can directly reference one of these images by using the `image` property in `.devcontainer/devcontainer.json` or updating the `FROM` statement in your own `Dockerfile` with one of the following:
```

src/python/README.md
dpprdan: Well, I wasn't aware that some of the build instructions (could) happen in devcontainer.json before @samruddhikhandale explained it in #653. If you're used to devcontainers this is obvious, I guess, but not necessarily if you're not IMHO. So at the risk of adding this for an audience of one (i.e. me), I'd say it might not be perfect, but still better ...? 

I am curious: Where are other tools and deps for this image defined, apart from `.devcontainer/devcontainer.json` and `.devcontainer/Dockerfile`.

bamurtaugh: Thanks for the great context here - it's always super helpful hearing where things are unclear or could be better.

What about mentioning the .devcontainer folder in general, rather than .devcontainer/devcontainer.json? That way it indicates the config is contained in essentially any files in this folder (which could be devcontainer.json, Dockerfile, Docker Compose).

src/python/README.md
bamurtaugh: This is great feedback. I think we'd still want to say "pre-built," but maybe we could link to our current docs content on pre-builds for more info? https://containers.dev/implementors/reference/#prebuilding

We also have an issue where @craiglpeters and I are going to work on highlighting and explaining pre-builds more clearly: https://github.com/devcontainers/spec/issues/261.

dpprdan: Yes, this makes much more sense. So we do the following?

```suggestion
You can directly reference [pre-built](https://containers.dev/implementors/reference/#prebuilding) versions of this image by using the `image` property in `.devcontainer/devcontainer.json` or updating the `FROM` statement in your own `Dockerfile` with one of the following:
```

src/python/README.md
dpprdan: It looses a bit of the (my) 🤯 of "the devcontainer.json (also) defines what's in the image" but this is indeed not the right place for this. So yeah, let's do 

```suggestion
Beyond Python and `git`, this image includes a number of Python tools, `zsh`, [Oh My Zsh!](https://ohmyz.sh/), a non-root `vscode` user with `sudo` access, and a set of common dependencies for development, see [.devcontainer](.devcontainer).
```

if that's still fine with you!

bamurtaugh: Looks great, thank you! As a small nit, this might flow a little more clearly for me:

```suggestion
Beyond Python and `git`, this image includes a number of Python tools, `zsh`, [Oh My Zsh!](https://ohmyz.sh/), a non-root `vscode` user with `sudo` access, and a set of common dependencies for development; see [.devcontainer/devcontainer.json](.devcontainer/devcontainer.json).
```

or

```suggestion
Beyond Python and `git`, this image includes a number of Python tools, `zsh`, [Oh My Zsh!](https://ohmyz.sh/), a non-root `vscode` user with `sudo` access, and a set of common dependencies for development. See [.devcontainer/devcontainer.json](.devcontainer/devcontainer.json) for the full config.
```

src/python/README.md
bamurtaugh: LGTM!

src/python/README.md
dpprdan: You probably mean

```suggestion
Beyond Python and `git`, this image includes a number of Python tools, `zsh`, [Oh My Zsh!](https://ohmyz.sh/), a non-root `vscode` user with `sudo` access, and a set of common dependencies for development. See [.devcontainer](.devcontainer) for the full config.
```

i.e. the link to the folder?! 😄

bamurtaugh: Haha yes, going too fast yesterday 😄

EMMANUEL2811UNITEC (approved):

dpprdan:
@bamurtaugh How would like to proceed here? Can this PR be merged?

bamurtaugh (approved):
Thanks for your contribution and collaboration!

bamurtaugh:
@samruddhikhandale are you okay with approving the workflow run here? Since that's more your area of expertise, I'll hold off.

src/python/README.md
```diff
@@ -38,13 +38,14 @@ You can decide how often you want updates by referencing a [semantic version](ht
 - `mcr.microsoft.com/devcontainers/python:1.0-3.9` (or `1.0-3.9-bullseye`, `1.0-3.9-buster`)
 - `mcr.microsoft.com/devcontainers/python:1.0.0-3.9` (or `1.0.0-3.9-bullseye`, `1.0.0-3.9-buster`)
 
-However, we only do security patching on the latest [non-breaking, in support](https://github.com/devcontainers/images/issues/90) versions of images (e.g. `1-14`). You may want to run `apt-get update && apt-get upgrade` in your Dockerfile if you lock to a more specific version to at least pick up OS security updates.
+However, we only do security patching on the latest [non-breaking, in support](https://github.com/devcontainers/images/issues/90) versions of images (e.g. `1-14`).
```
samruddhikhandale, line 41:
```suggestion
However, we only do security patching on the latest [non-breaking, in support](https://github.com/devcontainers/images/issues/90) versions of images (e.g. `1-3`). 
```

samruddhikhandale (approved):
Apologies for a late review, looks like this PR slipped through the gaps.

Thanks @bamurtaugh for looping me in. Overall the PR looks great, left one minor comment.

samruddhikhandale (approved):

PR #677, opened 2023-07-26
[javascript-node & typescript-node] Bump `npm` version due to GHSA-c2qf-rxjj-qqgw

alexander-smolyakov:
**Dev container name**: 

- javascript-node;
- typescript-node.

**Description**:

This PR addresses the GHSA-c2qf-rxjj-qqgw vulnerability. This vulnerability comes from the `semver` package, which is used as a dependency in the `npm`. To address the vulnerability, we need to bump the `npm` version up to [v9.8.1](https://github.com/npm/cli/releases/tag/v9.8.1).

*Changelog*:

- `Dockerfile` updated to bump `npm` version: `v9.5.1 -> v9.8.1`  

**Checklist**:

### DIFF
--- a/src/javascript-node/.devcontainer/Dockerfile
+++ b/src/javascript-node/.devcontainer/Dockerfile
@@ -8,9 +8,6 @@ ARG NPM_GLOBAL=/usr/local/share/npm-global
 # Add NPM global to PATH.
 ENV PATH=${NPM_GLOBAL}/bin:${PATH}
 
-# [Temporal] Bump npm version due to GHSA-c2qf-rxjj-qqgw
-RUN npm install -g npm@9.8.1
-
 RUN \
     # Configure global npm install location, use group to adapt to UID/GID changes
     if ! cat /etc/group | grep -e "^npm:" > /dev/null 2>&1; then groupadd -r npm; fi \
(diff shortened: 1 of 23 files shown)


src/javascript-node/.devcontainer/Dockerfile
```diff
@@ -8,6 +8,9 @@ ARG NPM_GLOBAL=/usr/local/share/npm-global
 # Add NPM global to PATH.
 ENV PATH=${NPM_GLOBAL}/bin:${PATH}
 
+# [Temporal] Bump npm version due to GHSA-c2qf-rxjj-qqgw
+RUN npm install -g npm@9.8.1
```
samruddhikhandale, line 12: Can we add a test/check to validate this change? Thanks!

alexander-smolyakov: Added test in https://github.com/devcontainers/images/pull/677/commits/697b9412efb01a456d5970aa28b8487587ae849b

samruddhikhandale: Hmm, I see a test added for checking the `npm` version, could we also add a check for the `semver` version? It would be nice to validate that the CVE is addressed.

PR #690, opened 2023-08-01
[anaconda] Address GHSA-47fc-vmwq-366v, GHSA-282v-666c-3fvg, GHSA-mrwq-x4v8-fh7p vulnerabilities

alexander-smolyakov:
**Dev container name**: 

- anaconda

**Description**:

This PR patches the following vulnerabilities:

- GHSA-47fc-vmwq-366v - related to the `torch` package;
- GHSA-282v-666c-3fvg - related to the `transformers` package;
- GHSA-mrwq-x4v8-fh7p - related to the `pygments` package;

These vulnerabilities come from the `continuumio/anaconda3` image used upstream for the anaconda devcontainer.

*Changelog*:

- Updated Dockerfile:
  - Removed lock from `continuumio/anaconda3` image to use the latest version of the image (Contains fixes for `torch` and `pygments` packages);
  - Locked versions for patched Python packages;
  - Added patch to install updated versions of `transformers` packages;

- Added tests to verify minimum versions of the following packages:
  - `torch` - _minimum package version set to `1.13.1`_;
  - `transformers` - _minimum package version set to `4.30.0`_;
  - `pygments` - _minimum package version set to `2.15.1`_;

- Updated information about packages in the devcontainer manifest;

- Updated tests to use different environments when installing packages from the `conda-forge` channel;

- Updated `README.md` to add info about possible conflicts in Conda's environment when channels are mixed;  

**Checklist**:


src/anaconda/.devcontainer/Dockerfile
```diff
@@ -18,15 +18,21 @@ RUN python3 -m pip install \
...
 
 RUN conda install \ 
     # https://github.com/advisories/GHSA-5cpq-8wj7-hf2v
     pyopenssl=23.2.0 \ 
     cryptography=41.0.2 \
     # https://github.com/advisories/GHSA-j8r2-6x86-q33q
-    requests=2.31.0
+    requests=2.31.0 \
+    # https://github.com/advisories/GHSA-mrwq-x4v8-fh7p
+    pygments=2.15.1
```
samruddhikhandale, line 35: From https://github.com/advisories/GHSA-mrwq-x4v8-fh7p, shouldn't it be `2.15.0` ?

samruddhikhandale: There's no harm in upgrading to latest, however, not sure how long the upstream would take to upgrade their util. Hence, would be good to pin to the min required version so that we could hope to remove this patch sooner.

src/anaconda/.devcontainer/Dockerfile
```diff
@@ -18,15 +18,21 @@ RUN python3 -m pip install \
     werkzeug \
     # https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-32862
     nbconvert \
-   # https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-28370
-    tornado
+    # https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-28370
+    tornado \
```
samruddhikhandale, line 22: While we are at it, should we pin all other utils to the current versions available in prod? The history files should provide this info. Else, it would be very difficult to sync with upstream and remove these patches in the future (without the need of degrading versions)

src/anaconda/.devcontainer/Dockerfile
alexander-smolyakov: > From https://github.com/advisories/GHSA-mrwq-x4v8-fh7p, shouldn't it be 2.15.0 ?

Only the `2.15.1` version is available in conda channel: https://anaconda.org/anaconda/pygments/files

src/anaconda/.devcontainer/Dockerfile
alexander-smolyakov: Sure, I will pin versions for all packages.

alexander-smolyakov: Resolved in a5bac2059541b2ee5954a3aa66ecb15ed958202f

src/anaconda/README.md
```diff
@@ -66,6 +66,14 @@ Given JavaScript front-end web client code written for use in conjunction with a
 }
 ```
 
+#### Using different Conda channels
+
+This devcontainer is based on the `ContinuumIO/anaconda3` docker image, which has all the `anaconda3` packages from defaults installed into the so-called base Conda's environment. It is recommended not to install packages from different channels in one environment since it could cause conflicts. When installing a package from a different channel (e.g., `conda-forge`) is required, the better approach is to create a new Conda environment.
```
samruddhikhandale, line 71:
```suggestion
This image is based on the `ContinuumIO/anaconda3` docker image, which has all the `anaconda3` packages from defaults installed into the base conda's environment. It is recommended not to install packages from different channels in one environment since it could cause conflicts. When installing a package from a different channel (e.g., `conda-forge`) is required, the better approach is to create a new conda environment.
```

src/anaconda/manifest.json
```diff
@@ -29,16 +29,15 @@
 			"cryptography",
 			"mistune",
 			"numpy",
-			"certifi",
-			"setuptools",
```
samruddhikhandale, line 33: Curious, why are we removing these utils? Are they no longer installed in the image?

alexander-smolyakov: We removed patches for these packages in the following PR: 
- #569

Related diff: [Code](https://github.com/devcontainers/images/commit/1f4139d3a9ac24154a1faf78790d8711007f43e4#diff-7b48baf632cdf4c52bc201952371ff2c6f43e0c6c129ecf20bf28f11a330bc76L67-L71)

These packages are still shipped with the devcontainer. Should I revert these changes?

samruddhikhandale: > These packages are still shipped with the devcontainer. Should I revert these changes?

Yep, it's nice to see the version history of these utils and it would prove handy if new CVEs are re-populated for them.

alexander-smolyakov: Got it! Reverted these changes in a64f9710d3fcec360271c1ab213605c8c2847ce5

samruddhikhandale (approved):

PR #692, opened 2023-08-02
[anaconda] Refactor `Dockerfile` to decrease image size

alexander-smolyakov:
**Description**:

During patch development, the CI checks for the `anaconda` devcontainer started to fail due to the space limit on a hosted agent. 

After investigation, it turned out that layer where we set up ownership and permission for the `opt` folder consumes a lot of space. This is related to using of `chown` and `chmod` commands.

_Related code_:

https://github.com/devcontainers/images/blob/47bbd234b8ecd36cd43c5bb9b36d986f19cc929c/src/anaconda/.devcontainer/Dockerfile#L88-L92

**Fix**:

To fix this and reduce image size, the Docker file was refactored:

- The `pip install` and `conda install` commands now executes against the `continuumio/anaconda3` image. This will allow us to not worry about `opt` folder permissions and ownership since we can set them up during `COPY` instruction;

- The `COPY` instruction [now sets up permissions](https://docs.docker.com/engine/reference/builder/#copy) and ownership for the `opt` folder. The `COPY` instruction will not create an additional layer, which should help decrease image size;

- Previous commands for setting up permissions and ownership were removed.


src/anaconda/.devcontainer/Dockerfile
```diff
@@ -3,9 +3,43 @@ FROM continuumio/anaconda3:2023.03-1 as upstream
...
+
+RUN conda install \ 
+    # https://github.com/advisories/GHSA-5cpq-8wj7-hf2v
+    pyopenssl=23.2.0 \ 
+    cryptography=41.0.2 \
+    # https://github.com/advisories/GHSA-j8r2-6x86-q33q
+    requests=2.31.0
+
 # Reset and copy updated files with updated privs to keep image size down
 FROM mcr.microsoft.com/devcontainers/base:0-bullseye
```
samruddhikhandale, line 32: Looks like we missed bumping the major version, can we update this? Thanks! 

```suggestion
FROM mcr.microsoft.com/devcontainers/base:1-bullseye
```

src/anaconda/.devcontainer/Dockerfile
```diff
@@ -3,9 +3,43 @@ FROM continuumio/anaconda3:2023.03-1 as upstream
...
-COPY --from=upstream /opt /opt/
+
+ARG USERNAME=vscode
+
+# Create the conda group and add remote user to the group
+RUN groupadd -r conda --gid 900 \ 
+    && usermod -aG conda ${USERNAME}
+
+# Copy opt folder, set ownership and group permissions
+COPY --chown=:conda --chmod=070 --from=upstream /opt /opt/
```
samruddhikhandale, line 41: Comparing this image changes with the prod image, we are missing some permissions on `/opt/conda`. Can we sync that? thanks! 

image

samruddhikhandale (changes requested):
It is super cool to see the image size decrease from ~9 GB to ~5 GB 🎉 ✨ 
Thank you! 

Left some comments!

src/anaconda/.devcontainer/Dockerfile
alexander-smolyakov: I bumped container version in 6e511af45af5666dbda0bda29502efaab7c86fce

src/anaconda/.devcontainer/Dockerfile
alexander-smolyakov: I updated permissions in 6e511af45af5666dbda0bda29502efaab7c86fce

src/anaconda/.devcontainer/Dockerfile
```diff
@@ -3,9 +3,43 @@ FROM continuumio/anaconda3:2023.03-1 as upstream
...
+FROM mcr.microsoft.com/devcontainers/base:1-bullseye
+
+ARG USERNAME=vscode
+
+# Create the conda group and add remote user to the group
+RUN groupadd -r conda --gid 900 \ 
+    && usermod -aG conda ${USERNAME}
+
+# Copy opt folder, set ownership and group permissions
+COPY --chown=:conda --chmod=775 --from=upstream /opt /opt/
```
samruddhikhandale, line 41:
```suggestion
COPY --chown=:conda --chmod=775 --from=upstream /opt/conda /opt/conda
```

> && chown -R :conda /opt/conda \
    && chmod -R g+w /opt/conda \
    
Syncing with ^

samruddhikhandale: I agree that the current upstream image only adds `conda` folder to `/opt/`, however, that could change anytime. Hence, it would be nice to be as precise as possible!

samruddhikhandale (approved):

src/anaconda/.devcontainer/Dockerfile
alexander-smolyakov: Good point! I applied the suggested changes.

alexander-smolyakov: Related changes caused the permission denied error. I'm investigating why this happened.
Error: https://github.com/devcontainers/images/actions/runs/5760266653/job/15615884009#step:3:5724

alexander-smolyakov: The issue is related to the fact that when using chown and chmod functions with the `COPY` command, Docker updates the metadata of folder content but not the folder itself. I updated Dockerfile to set permissions to the `opt/conda` folder: https://github.com/devcontainers/images/pull/692/commits/ffece92eedb493a66f2cd3798efecbfa29c7a734.

Results:

> COPY /opt/conda /opt/conda # buildkit

```console
Permission     UID:GID       Size  Filetree
drwxr-xr-x         0:0     4.2 GB  └── opt 
drwxr-xr-x       0:900     4.2 GB      └── conda
```

> COPY /opt/ /opt/ # buildkit

```console
Permission     UID:GID       Size  Filetree
drwxr-xr-x         0:0     4.2 GB  └── opt
drwxrwxr-x       0:900     4.2 GB      └── conda
```

> COPY /opt/conda /opt/conda # buildkit
> RUN |1 USERNAME=vscode /bin/sh -c chmod =2775 /opt/conda # buildkit

```console
Permission     UID:GID       Size  Filetree
drwxr-xr-x         0:0        0 B  └── opt
drwxrwsr-x       0:900        0 B      └── conda
```

samruddhikhandale: Nice, thanks!

PR #741, opened 2023-09-11
[miniconda] Refactor Dockerfile to sync with `anaconda` devcontianer

alexander-smolyakov:
**Devcontainer name**: 

- miniconda

**Description**:

This PR contains the same changes as in #692. The main idea is to make the anaconda and miniconda devcontainers consistent.

*Changelog*:

- The `pip install` and `conda install` commands now execute against the `continuumio/miniconda3` image. (_This will allow us to not worry about `opt` folder permissions and ownership since we can set them up during `COPY` instruction_);

- The `COPY` instruction [now sets up permissions](https://docs.docker.com/engine/reference/builder/#copy) and ownership for the `opt` folder. The `COPY` instruction will not create an additional layer, which should help decrease image size;

- Previous commands for setting up permissions and ownership were removed.

**Checklist**:

### DIFF
--- a/src/miniconda/.devcontainer/Dockerfile
+++ b/src/miniconda/.devcontainer/Dockerfile
@@ -1,10 +1,30 @@
 FROM continuumio/miniconda3 as upstream
 
+# Temporary: Upgrade python packages due to mentioned CVEs
+# They are installed by the base image (continuumio/miniconda3) which does not have the patch.
+RUN conda install \ 
+    # https://github.com/pyca/cryptography/security/advisories/GHSA-5cpq-8wj7-hf2v
+    pyopenssl=23.2.0 \ 
+    cryptography=41.0.2 \
+    # https://github.com/advisories/GHSA-j8r2-6x86-q33q
+    requests=2.31.0
+
 # Reset and copy updated files with updated privs to keep image size down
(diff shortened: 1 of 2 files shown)


src/miniconda/.devcontainer/Dockerfile
```diff
@@ -1,10 +1,30 @@
...
+# They are installed by the base image (continuumio/miniconda3) which does not have the patch.
+RUN conda install \ 
+    # https://github.com/pyca/cryptography/security/advisories/GHSA-5cpq-8wj7-hf2v
+    pyopenssl=23.2.0 \ 
+    cryptography=41.0.2 \
+    # https://github.com/advisories/GHSA-j8r2-6x86-q33q
+    requests=2.31.0
+
 # Reset and copy updated files with updated privs to keep image size down
 FROM mcr.microsoft.com/devcontainers/base:0-bullseye
```
samruddhikhandale, line 13:
```suggestion
FROM mcr.microsoft.com/devcontainers/base:1-bullseye
```

Looks like we missed updating this image!

samruddhikhandale (approved):
Looks good, left one comment!

samruddhikhandale (approved):

PR #753, opened 2023-09-13
[universal] Address GHSA-jm77-qphf-c4w8 and GHSA-r9hx-vwmv-q579 vulnerabilities 

alexander-smolyakov:
**Devcontainer name**: 

- universal

**Description**:

This PR patches the following vulnerabilities:

- GHSA-jm77-qphf-c4w8 - related to the `cryptography` package;
- GHSA-r9hx-vwmv-q579 - related to the `setuptools` package. The vulnerability was patched a while ago, but the `site-packages` folder contained an older version of the related package, which triggers security scanners;

**Changelog**:

`patch-conda` feature:

- Bumped `cryptography` package version to address GHSA-jm77-qphf-c4w8;

`patch-python` feature:

- Removed the `--user` option from the `pip install` command. To make sure that the `site-packages` folder doesn't contain older versions of patched packages.
 
Tests:

- Updated test to verify `cryptography` minimum version (_Minimum package version set to `41.0.3` which fixes GHSA-jm77-qphf-c4w8_);

- Updated tests to use different environments when installing packages from the `conda-forge` channel;

- Conda tests were renamed to make names more explicit;

**Checklist**:

### DIFF
--- a/src/universal/.devcontainer/local-features/patch-python/install.sh
+++ b/src/universal/.devcontainer/local-features/patch-python/install.sh
@@ -34,7 +34,7 @@ update_package() {
     PACKAGE=$2
 
     sudo_if "$PYTHON_PATH -m pip uninstall --yes $PACKAGE"
-    sudo_if "$PYTHON_PATH -m pip install --user --upgrade --no-cache-dir $PACKAGE"
+    sudo_if "$PYTHON_PATH -m pip install --upgrade --no-cache-dir $PACKAGE"
 }
 
 # Temporary: Upgrade python packages due to security vulnerabilities
--- a/src/universal/.devcontainer/local-features/patch-conda/install.sh
+++ b/src/universal/.devcontainer/local-features/patch-conda/install.sh
@@ -50,9 +50,10 @@ sudo_if /opt/conda/bin/python3 -m pip install --upgrade pip
 # Temporary: Upgrade python packages due to security vulnerabilities
 # They are installed by the conda feature and Conda distribution does not have the patches.
 
-# https://github.com/advisories/GHSA-5cpq-8wj7-hf2v
+# pyopenssl should be updated to be compatible with latest version of cryptography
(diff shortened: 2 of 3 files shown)


src/universal/.devcontainer/local-features/patch-python/install.sh
```diff
@@ -34,7 +34,7 @@ update_package() {
     PACKAGE=$2
 
     sudo_if "$PYTHON_PATH -m pip uninstall --yes $PACKAGE"
-    sudo_if "$PYTHON_PATH -m pip install --user --upgrade --no-cache-dir $PACKAGE"
+    sudo_if "$PYTHON_PATH -m pip install --upgrade --no-cache-dir $PACKAGE"
```
samruddhikhandale, line 37: why do we need to remove `--user` flag? Would this be a breaking change for users?

alexander-smolyakov: > why do we need to remove --user flag? 

When the `--user` flag is used, the PIP installs the package to the local user folder. In our case, we want to patch packages presented in the `site-packages` folder. Current behavior results in having a patched version of the package in the user folder and an old version in the `site-packages` folder. Additionally, in this case, our security scanners fire alerts because of the old version.

> Would this be a breaking change for users?

This feature was implemented for internal purposes, and I don't see breaking changes since we didn't change the installed versions of these packages.

samruddhikhandale: I see, one thing to note is that https://github.com/devcontainers/features/blob/main/src/python/install.sh#L313 Feature installs packages using `--user` flag. Hence, we need to be able to distinguish and upgrade packages accordingly (ie. patches from upstream image vs Feature). I believe the Feature always installs the latest packages, but thought of pointing this out.

samruddhikhandale (approved):

samruddhikhandale (approved):

PR #759, opened 2023-09-20
New workflow - Allows manually pushing dev images from branch

samruddhikhandale:
It will be helpful to test if changes to a branch works in an image.

Scenario - Universal image takes a while to be built, so it will be helpful to make changes to a branch and test them with an image (without merging the PR). We could use actions and tests to validate our changes but at times actions don't replicate the scenario. (eg. universal image when built with latest Feature version of dnd Feature fails to start docker within a codespace but passes in an action. This new workflow will be helpful in such cases)

Note: This workflow can only push dev tags, prod tags cannot be pushed by this workflow.

### DIFF
--- a/.github/workflows/push-manual-dev.yml
+++ b/.github/workflows/push-manual-dev.yml
@@ -1,12 +1,8 @@
-name: Push dev-image manually
+name: Re-push image
 
 on:
   workflow_dispatch:
     inputs:
-      release:
-        description: 'Release branch'
-        required: true
-        default: 'main'
       image:
         description: 'Image ID'
         required: true
@@ -73,7 +69,7 @@ jobs:
         ln -s "$GITHUB_WORKSPACE/ref/node_modules" node_modules
 
         build/vscdc push  --replace-images \
-                          --release ${{ github.event.inputs.release }} \
+                          --release main \
                           --registry "$REGISTRY" \
                           --registry-path "$REGISTRY_BASE_PATH" \
                           --stub-registry "$STUB_REGISTRY" \
--- a/.github/workflows/push-manual-dev.yml
+++ b/.github/workflows/push-manual-dev.yml
@@ -1,0 +1,82 @@
+name: Push dev-image manually
+
+on:
+  workflow_dispatch:
+    inputs:
+      release:
+        description: 'Release branch'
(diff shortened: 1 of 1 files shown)


.github/workflows/push-manual-dev.yml
```diff
@@ -0,0 +1,82 @@
+name: Push dev-image manually
+
+on:
+  workflow_dispatch:
+    inputs:
+      release:
+        description: 'Release branch'
```
joshspicer, line 7: I think `workflow_dispatch` gives you this for free (see https://github.com/devcontainers/features/actions/workflows/test-manual.yaml)?

.github/workflows/push-manual-dev.yml
```diff
@@ -0,0 +1,82 @@
...
+        npm install -g @devcontainers/cli
+
+        # Go to the release, symlink the build tool from ref since this is the version for the workflow
+        cd "$GITHUB_WORKSPACE/release"
+        rm -rf build node_modules
+        ln -s "$GITHUB_WORKSPACE/ref/build" build
+        ln -s "$GITHUB_WORKSPACE/ref/node_modules" node_modules
+
+        build/vscdc push  --replace-images \
+                          --release ${{ github.event.inputs.release }} \
```
joshspicer, line 76: Is the value of `--release` what determines whether it pushes dev tags or real tags?  What if this pipeline is run on a tag (which workflow_dispatch would be ok doing)?

joshspicer (commented):
Added a couple questions!

.github/workflows/push-manual-dev.yml
samruddhikhandale: Good point, removed!

.github/workflows/push-manual-dev.yml
samruddhikhandale: We have 👇 which skips the run if the user selects a tag, however, updated it to `main` to ensure we always push dev tags.

https://github.com/devcontainers/images/pull/759/files#diff-5954ba0f683a9e24316be62ceac8e931df97839f38c87eafc513c71e04b8a753R13

joshspicer (approved):

PR #774, opened 2023-09-29
v0.3.20

samruddhikhandale:
Changelog

- [[anaconda] Update `jupyter_server` package due to GHSA-r726-vmfq-j9j3](https://github.com/devcontainers/images/pull/754)
- [[Typesrcipt-node][Javascript-node] EOL Node 16](https://github.com/devcontainers/images/pull/751)
- [[universal] Address GHSA-jm77-qphf-c4w8 and GHSA-r9hx-vwmv-q579 vulnerabilities](https://github.com/devcontainers/images/pull/753)
- [Support: Java 21](https://github.com/devcontainers/images/pull/770)

### DIFF
--- a/build/config.json
+++ b/build/config.json
@@ -226,7 +226,7 @@
 			"downloadUrl":"https://xdebug.org/"
 		},
 		"Java": {
-			"versionCommand": "java --version | grep -oP 'openjdk\\s+\\K[0-9]+\\.[0-9]+\\.[0-9]+'",
+			"versionCommand": "java --version | grep -oP 'openjdk\\s+\\K[0-9]+\\.[0-9]+\\.[0-9]+|[0-9]{2}' | head -n 1",
 			"downloadUrl": "https://adoptopenjdk.net/"
 		},
 		".NET": {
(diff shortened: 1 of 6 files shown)


build/config.json
```diff
@@ -226,7 +226,7 @@
 			"downloadUrl":"https://xdebug.org/"
 		},
 		"Java": {
-			"versionCommand": "java --version | grep -oP 'openjdk\\s+\\K[0-9]+\\.[0-9]+\\.[0-9]+'",
+			"versionCommand": "java --version | grep -oP 'openjdk\\s+\\K[0-9]+\\.[0-9]+\\.[0-9]+|[0-9]{2}' | head -n 1",
```
samruddhikhandale, line 229: Fixes https://github.com/devcontainers/images/actions/runs/6341911076/job/17262760053#step:5:4421

joshspicer: wow that's quite the regex.  I wonder if for some of these we could split into another file and run tests just against them 😓 . Might save us from some bugs in the future, and wouldn't slow down the workflow runs vs trying EVERY single combo.  Something for us to think about :D

joshspicer (approved):

build/config.json
samruddhikhandale: Yes, I definitely agree 👍 These errors happen rarely, but definitely a pain. 
Generating version history command can definitely use some more ❤️

PR #795, opened 2023-10-06
[universal] Update `cryptography` package due to GHSA-v8gr-m533-ghj9

alexander-smolyakov:
**Devcontainer name**: 

- universal

**Description**:

This PR addresses the GHSA-v8gr-m533-ghj9 vulnerability. The vulnerability comes from the `conda` distribution and is related to the `cryptography` package.

*Changelog*:

- Updated `patch-conda` feature to install the latest `cryptography` package version;

- Updated test to verify `cryptography` minimum version (_Minimum package version set to `41.0.4` which fixes GHSA-v8gr-m533-ghj9_);

- Removed check for the minimum version of "wheel" since the package is no longer present in the PIP list;

**Checklist**:


samruddhikhandale (approved):
Thanks ✨

PR #810, opened 2023-10-17
[Cpp] - Only exit vcpkg install early on Buster ARM64

tarebyte:
I ran into an issue where all ARM64 OS builds are blocked even if they ship with CMake > 3.15 This appears to be a regression from https://github.com/devcontainers/images/pull/588.

Looking through the packages I think that the only OS that doesn't ship with the right CMake is [Debian Buster](https://packages.debian.org/buster/cmake).

This pull request updates the conditional to add back in the denylist so that other new operating systems can install `vcpkg`.

~~I've also updated the script logic so that `VCPKG_FORCE_SYSTEM_BINARIES` is set when using `arm64`. If there is a better way to do this please let me know!~~

Thanks!


tarebyte:

src/cpp/.devcontainer/base-scripts/install-vcpkg.sh
```diff
@@ -10,10 +10,17 @@ USERNAME=${1:-"vscode"}
 
 . /etc/os-release
 
-# Exit early if ARM64 OS does not have cmake version required to build Vcpkg
+# The buster pkg repo install cmake version < 3.15 which is required to run bootstrap-vcpkg.sh on ARM64
+VCPKG_UNSUPPORTED_ARM64_VERSION_CODENAMES="buster"
```
samruddhikhandale, line 14: Thank you fixing the bug, apologies for the regression!

src/cpp/.devcontainer/base-scripts/install-vcpkg.sh
```diff
@@ -80,6 +87,13 @@ if [[ "\${PATH}" != *"\${VCPKG_ROOT}"* ]]; then export PATH="\${PATH}:\${VCPKG_R
 EOF
 )"
 
+if [[ -n "$VCPKG_FORCE_SYSTEM_BINARIES" ]]; then
+    updaterc "$(cat << EOF
+export VCPKG_FORCE_SYSTEM_BINARIES=1
```
samruddhikhandale, line 92: I understand how `VCPKG_FORCE_SYSTEM_BINARIES` can be useful in some cases to speed up the build process and conserve resources. 

Few questions: 

1. Why should we only set it for ARM architectures and not for others? 
2. By default, vcpkg builds packages from source and it is generally recommended to build packages from source whenever possible to ensure compatibility and stability with the target system. Hence, should we let the user make this decision? (as there are benefits and drawbacks of using system-provided binaries and let user make an informed decision based on their specific requirements and constraint)

Super interested in understanding your reasoning behind this change. We can also open a new issue and gather interest/feedback from the community for setting this by default and reiterate on this decision if necessary, thanks!

samruddhikhandale (commented):
Thank you for taking the time to contribute this PR, left some clarifying questions!

src/cpp/.devcontainer/base-scripts/install-vcpkg.sh
tarebyte: > Why should we only set it for ARM architectures and not for others?

Great question, when I read it the first time it looked like the script was only concerned with `arm`.

According to the error message I ran into, it looks like the script should account for `arm`, `s390x`, `ppc64le`, and `riscv`.

```
Environment variable VCPKG_FORCE_SYSTEM_BINARIES must be set on arm, s390x, ppc64le and riscv platforms.
```

I can attempt to make that change. My goal was to only set `VCPKG_FORCE_SYSTEM_BINARIES` when necessary.

thsmdt:
Hey @tarebyte -- are you still working on this? :)

I am currently running into this regression as well, where I am missing `vcpkg` in the arm-images, and was wondering if someone was already contributing a fix when I found your PR.

As far as I understand the modifications concerning `VCPKG_FORCE_SYSTEM_BINARIES` are optional. Would it be okay for you to remove them for now and as @samruddhikhandale recommended have a separate issue for it?

samruddhikhandale (approved):
Looks good to me, thanks!

PR #812, opened 2023-10-17
v0.3.23

samruddhikhandale:
Changelog -

- [[universal] Update cryptography package due to GHSA-v8gr-m533-ghj9](https://github.com/devcontainers/images/pull/795)
- [[anaconda] Address vulnerabilities: GHSA-j7hp-h8jx-5ppr, GHSA-v845-jxx5-vc9f](https://github.com/devcontainers/images/pull/801)
- [[miniconda] Update urllib3 package due to GHSA-v845-jxx5-vc9f](https://github.com/devcontainers/images/pull/802)
- [[universal] Update urllib3 package due to GHSA-v845-jxx5-vc9f](https://github.com/devcontainers/images/pull/803)
- [[Dotnet] - Supports bookworm (New major version)](https://github.com/devcontainers/images/pull/809)

PR #807, opened 2023-10-16
[anaconda] Update `imagecodecs` package due to GHSA-94vc-p8w7-5p49

alexander-smolyakov:
**Devcontainer name**: 

- anaconda

**Description**:

This PR addresses the GHSA-94vc-p8w7-5p49 vulnerability. The vulnerability comes from the `continuumio/anaconda3` image and is related to the `imagecodecs` package.

*Changelog*:

- Bumped `imagecodecs` package version to address GHSA-v845-jxx5-vc9f;

- Added test to verify `imagecodecs` minimum version (_Minimum package version set to `2023.9.18` which fixes GHSA-94vc-p8w7-5p49_);

**Checklist**:

- [x] Checked that applied changes work as expected

PR #815, opened 2023-10-18
[anaconda] Rework patch for GHSA-94vc-p8w7-5p49

alexander-smolyakov:
**Devcontainer name**: 

- anaconda

**Description**:

The previous patch (#807) was reverted in #812 due to the following error:

```console
ERROR: Could not build wheels for imagecodecs, which is required to install pyproject.toml-based projects
```

It seems the issue could be related to old versions of `setuptools` and `wheel` packages. To address this issue, the patch was reworked to update versions of `setuptools` and `wheel` packages along with `imagecodecs`.

*Changelog*:

- Updated `setuptools` and `wheel` to avoid potential issues with installing newer versions of packages;

- Bumped `imagecodecs` package version to address GHSA-94vc-p8w7-5p49;

- Added test to verify `imagecodecs` minimum version (_Minimum package version set to `2023.9.18` which fixes GHSA-94vc-p8w7-5p49_);

**Checklist**:

### DIFF
--- a/src/anaconda/.devcontainer/Dockerfile
+++ b/src/anaconda/.devcontainer/Dockerfile
@@ -5,9 +5,12 @@ RUN . /etc/os-release && if [ "${VERSION_CODENAME}" != "bullseye" ]; then exit 1
 
 # Temporary: Upgrade python packages due to mentioned CVEs
 # They are installed by the base image (continuumio/anaconda3) which does not have the patch.
-RUN conda install \ 
+RUN conda install \
+    # Update "setuptools" and "wheel" to avoid potential issues with installing newer versions of packages
+    setuptools=68.0.0 \
+    wheel=0.41.2 \
     # pyopenssl should be updated to be compatible with latest version of cryptography
-    pyopenssl=23.2.0 \ 
+    pyopenssl=23.2.0 \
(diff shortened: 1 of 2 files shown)


src/anaconda/.devcontainer/Dockerfile
```diff
@@ -5,9 +5,12 @@ RUN . /etc/os-release && if [ "${VERSION_CODENAME}" != "bullseye" ]; then exit 1
 
 # Temporary: Upgrade python packages due to mentioned CVEs
 # They are installed by the base image (continuumio/anaconda3) which does not have the patch.
-RUN conda install \ 
+RUN conda install \
+    # Update "setuptools" and "wheel" to avoid potential issues with installing newer versions of packages
+    setuptools=68.0.0 \
```
samruddhikhandale, line 10: Do we need to pin `setuptools` and `wheel` to this specific version for making it compatible with `imagecodecs` ? 
If so, can we add a note that `imagecodecs` is dependent on them, so that we don't unpin it accidentally?

samruddhikhandale:
Unfortunately, the image build is still failing with [these changes](https://github.com/devcontainers/images/compare/samruddhikhandale/test-rework-imagecodes)^ . See https://github.com/devcontainers/images/actions/runs/6565254242/job/17833552777#step:6:2240

PR #853, opened 2023-11-15
Drop dotnet:8.0-bullseye due to upstream unavailability

samruddhikhandale:
Fixes https://github.com/devcontainers/images/actions/runs/6882081495/job/18719819258#step:6:3594

Opened https://github.com/dotnet/dotnet-docker/issues/5001 as dotnet images for `8.0-bullseye` are unavailable.

### DIFF
--- a/src/dotnet/README.md
+++ b/src/dotnet/README.md
@@ -9,7 +9,7 @@
 | *Categories* | Core, Languages |
 | *Image type* | Dockerfile |
 | *Published images* | mcr.microsoft.com/devcontainers/dotnet |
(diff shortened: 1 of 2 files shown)


joshspicer (approved):

samruddhikhandale:
From https://github.com/dotnet/dotnet-docker/discussions/5002#discussioncomment-7581104, hence, we won't add support for `dotnet:8.0-bullseye`

> according to our [supported platforms documentation](https://github.com/dotnet/dotnet-docker/blob/main/documentation/supported-platforms.md#linux), we support the latest stable version of Debian at the time of a .NET major release as well as any new stable Debian versions as they are released. Since the latest stable version of Debian is Bookworm, we won't be offering Bullseye images for .NET 8.

PR #852, opened 2023-11-14
Adds support for .NET 8.0

samruddhikhandale:
Ref: https://github.com/devcontainers/images/issues/787

Issue #787, opened 2023-10-04
Support .NET 8 scheduled for Nov 14


samruddhikhandale:
Depends on [add a reference link here]

- [x] Open a PR 
- [x] Create a release
- [ ] Update [templates](https://github.com/devcontainers/templates) (if necessary)
- [x] Update https://github.com/devcontainers/images/issues/90
- [ ] Update [Microsoft Artifact Registry](https://mcr.microsoft.com/en-us/catalog?search=dev%20container) docs

marinasundstrom:
So there is no build with RC 2?

samruddhikhandale:
> So there is no build with RC 2?

@marinasundstrom, unfortunately, we don't. We only add support for production ready frameworks.

seesharprun:
Would it be out-of-bounds to create a PR to add .NET 8 support to the universal image as it's now the new LTS?

samruddhikhandale:
> Would it be out-of-bounds to create a PR to add .NET 8 support to the universal image as it's now the new LTS?

For the `universal` image, we have the following rules defined :

- Only include LTS (or latest) releases of tools used for key scenarios
- When several LTS versions are in support, include only the latest two

Hence, updating `universal` image to support .NET versions `8` and `6` makes sense to me. 

I can help review [Update universal image to .NET 8 #855](https://github.com/devcontainers/images/pull/855) when it's ready.

seesharprun:
Sounds good, I removed .NET 7 from the pull request and it's ready for review now.

Thanks!

samruddhikhandale:
Closing as .Net 8 images are now released 🎉

PR #855, opened 2023-11-16
Update universal image to .NET 8

seesharprun:
.NET 8 is now the LTS version of .NET: https://dotnet.microsoft.com/platform/support/policy/dotnet-core#lifecycle

This PR proposes to update the universal image to install .NET 8 along with 7 and 6.

- [x] Update .NET feature to version 2
- [x] Install .NET 8 as default version
- [x] Install .NET 6 as a side-by-side verison
- [x] Remove `installUsingApt` option as it's not available anymore for the feature: https://github.com/devcontainers/features/blob/main/src/dotnet/devcontainer-feature.json#L7-L36


seesharprun:
- Related to #852
- Related to #787

Still a WIP

seesharprun:
I'm not familiar with Oryx enough to figure out why it's failing to build in the smoke test.

seesharprun:
Here's the error message I'm seeing in the smoke test:

```
33.67 Build FAILED.
33.67 
33.67 /opt/tmp/oryx-repo/src/BuildScriptGenerator.Common/BuildScriptGenerator.Common.csproj : error NU1903: Warning As Error: Package 'Azure.Identity' 1.8.2 has a known high severity vulnerability, https://github.com/advisories/GHSA-5mfx-4wcx-rv27 [/opt/tmp/oryx-repo/Oryx.sln]
33.67     0 Warning(s)
33.67     1 Error(s)
33.67 
33.67 Time Elapsed 00:00:05.39
33.73 ERROR: Feature "Oryx" (ghcr.io/devcontainers/features/oryx) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/oryx for help troubleshooting this error.
```

If you look at the Oryx repo, they are indeed using version 1.8.2 of Azure.Identity:

https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator.Common/BuildScriptGenerator.Common.csproj#L19

If you look at NuGet, that version has a security vulnerability and that's being thrown as an error here and may be a warning elsewhere:

https://www.nuget.org/packages/Azure.Identity/1.8.2

samruddhikhandale:
Oryx tool needs to be built with .NET 7, however, looking at https://github.com/devcontainers/images/actions/runs/6895550673/job/18760553721?pr=855#step:3:18052, it's getting built with .NET 8. This is definitely a bug in the [oryx Feature](https://github.com/devcontainers/features/tree/main/src/oryx) which needs to be fixed to unblock this PR.

> If you look at NuGet, that version has a security vulnerability and that's being thrown as an error here and may be a warning elsewhere:

I'll reach out to the Oryx team and mention this security vulnerability. Thanks for pointing it out.
These build failures don't happen with .NET 7, hence, I wonder if .NET 8 is throwing them as errors as opposed to warnings.

samruddhikhandale:
> I'll reach out to the Oryx team and mention this security vulnerability. Thanks for pointing it out.

Fix - https://github.com/microsoft/Oryx/pull/2263 🎉

seesharprun:
@samruddhikhandale, did the PR from the Oryx team resolve this issue?

samruddhikhandale:
> did the PR from the Oryx team resolve this issue?

It did resolve the issue we were seeing in https://github.com/devcontainers/images/pull/855#issuecomment-1815345494

> Oryx tool needs to be built with .NET 7, however, looking at https://github.com/devcontainers/images/actions/runs/6895550673/job/18760553721?pr=855#step:3:18052, it's getting built with .NET 8. This is definitely a bug in the [oryx Feature](https://github.com/devcontainers/features/tree/main/src/oryx) which needs to be fixed to unblock this PR.

(From https://github.com/devcontainers/images/pull/855#issuecomment-1815369560)

However, this ^ is another issue which is currently blocking this PR. I have opened https://github.com/devcontainers/features/issues/760 for tracking, but have been super busy and wasn't able to get to it. Apologies for that and  thanks for your patience. I am hoping to fix https://github.com/devcontainers/features/issues/760 sometime this week.

However, in case you are interested in contributing a fix for https://github.com/devcontainers/features/issues/760, we'd really appreciate that. ✨

seesharprun:
I missed that follow-up feature.

It looks like the Oryx feature auto-builds with the primary (latest?) version of .NET and we're only putting .NET 6 & 8 on the image.

Forgive me if this seems like an obvious question, but how should this be resolved:

- Should the image be updated to have .NET 6, 7, and 8 and the feature updated to specify .NET 7?
- Should the feature be updated to install .NET 7 specifically in isolation and use that to install the feature?
- Should Oryx be updated to .NET 8?

Sorry, if you can't answer those questions. Oryx is still a bit of a mystery to me as I only ever encountered it with Azure App Service. I'm happy to contribute if I can, but I don't have a ton of experience with that project.

samruddhikhandale:
No worries, and thank you for your willingness to help @seesharprun, appreciate it.

> Should Oryx be updated to .NET 8?

Currently, the oryx tool is built with .Net 7, they would switch it to .Net 8 sometime in the future, but I don't think it's worth waiting/ getting blocked because of it.

In the meanwhile, we can update the [Oryx Feature](https://github.com/devcontainers/features/tree/main/src/oryx) to do the following -
- Install .Net 7
- Build oryx binary
- Uninstall .Net 7

Ideally, the Oryx Feature should be doing that ^, see https://github.com/devcontainers/features/blob/main/src/oryx/install.sh#L136-L143 but maybe there is a bug? Previously, we had .Net  7 in the `universal` image, so most likely we didn't encounter it?

 > Should the image be updated to have .NET 6, 7, and 8 and the feature updated to specify .NET 7?
 
 Nope, I don't expect changes to the image code.
 
> Should the feature be updated to install .NET 7 specifically in isolation and use that to install the feature?

Yes, about right

src/universal/.devcontainer/devcontainer.json
```diff
@@ -9,10 +9,9 @@
             "userUid": "1000",
             "userGid": "1000"
         },
-        "ghcr.io/devcontainers/features/dotnet:1": {
-            "version": "7.0.306",
-            "installUsingApt": "false",
-            "additionalVersions": "6"
+        "ghcr.io/devcontainers/features/dotnet:2": {
```
samruddhikhandale, line 12:
```suggestion
        "ghcr.io/devcontainers/features/dotnet:1": {
```

Can you apply this one change and see if it fixes oryx build issues? We did major refactoring to the .Net Feature which also looks suspicious. This could also explain why oryx does not use .Net 7 from apt and prefers the one installed via Feature (I could be wrong as well)

> Update .NET feature to version 2

If v1 of the dotnet Feature helps us get this PR unblocked then we could work on this ^ in a separate PR/issue.

seesharprun:
Unfortunately, I can't test the image locally. I'm using Windows and it doesn't seem to let me install the local NVS feature.

seesharprun:
This error makes me assume I need bash locally to test out the devcontainer:

image

seesharprun:
In the interim, would adding **.NET 7** back solve the Oryx build error?

```json
"ghcr.io/devcontainers/features/dotnet:1": {
    "version": "6.0",
    "additionalVersions": "7.0, 8.0"
},
```

or

```json
"ghcr.io/devcontainers/features/dotnet:2": {
    "version": "6.0",
    "additionalVersions": "7.0, 8.0"
},
```

I'm going to change to the latter if you don't mind running the smoke test again.

seesharprun:
Another update, it looks like using the older version of the feature causes an error when installing multiple versions of .NET.

I'm attempting to smoke test on my end with the latest version of the feature:

```json
"ghcr.io/devcontainers/features/dotnet:1": {
    "version": "8.0",
    "additionalVersions": "7.0, 6.0"
},
```

For now, I will mark this PR as draft until there's a solution.

src/universal/.devcontainer/devcontainer.json
seesharprun: Switching back to `1` did not resolve the Oryx build issues. I switched back to `2` and then added .NET 7.

seesharprun:
I agree @samruddhikhandale. I'm just not entirely sure how to force Oryx to use .NET 7 during the build process.

seesharprun:
As a first step, I'm going to try a draft PR where I add the flag `--framework net7.0` to be more explicit about using .NET 7.0 in the three .NET CLI instructions:

https://github.com/devcontainers/features/blob/main/src/oryx/install.sh#L161-L164

samruddhikhandale:
// cc @eljog and @joshspicer for visibility.

glaucia86:
Hi, everyone! 
I'm dealing with the same issue.... once we have a huge workshop here: **[Generative A.I for Beginners](https://github.com/microsoft/generative-ai-for-beginners?WT.mc_id=academic-111247-leestott)** and we're including some dotnet code samples and using the Polyglot Jupyter extension; But this extension is asking for .NET 8 version

And once we're using the `mcr.microsoft.com/devcontainers/universal:2` image, we cannot run for example this sample on Codespaces: **[search application](https://github.com/microsoft/generative-ai-for-beginners/blob/main/08-building-search-applications/dotnet/notebook-azure-openai.dib)**

We still waiting for answers and solution for this issue!

src/universal/.devcontainer/devcontainer.json
```diff
@@ -9,10 +9,9 @@
...
+            "version": "8.0",
```
timheuer, line 13: IMO mark this as `8.0.x'

src/universal/.devcontainer/devcontainer.json
```diff
@@ -9,10 +9,9 @@
...
+            "additionalVersions": "7.0.306, 6.0"
```
timheuer, line 14: `additionalVersions` brings in additional SDKs.  From a product owner perspective, we would prefer not to do this as the most recent SDK can build projects targeting these versions. Of course those that have pinning (global.json) will need the specific SDK they want, but it is likely not going to serve everyone.

I'd recommend removing the v7 from additionalVersions here.  What would be helpful in the devcontainer scenario is adding the v7 runtime to `dotnetRuntimeVersions` for this entry as likely VSCode extensions that folks will use with this container are still targeting the runtime and this would help.  So I'd recommend removing this from SDK (additionalVersions) and instead putting `7.0.x` on `dotnetRuntimeVersions` attribute like we use on our official codespaces container: https://github.com/github/dotnet-codespaces/blob/main/.devcontainer/devcontainer.json#L18

seesharprun: @timheuer, what do you think about .NET 6? Should I include that in the additional runtimes?

I'm considering this:

```json
"ghcr.io/devcontainers/features/dotnet:2": {
    "version": "8.0.x",
    "dotnetRuntimeVersions": "7.0.x, 6.0.x",
    "aspNetCoreRuntimeVersions": "7.0, 6.0.x"
}
```

timheuer: I think that would be fine yes actually. (putting the .x on the aspnet one too to get patches).

timheuer:
> And since we're using the `mcr.microsoft.com/devcontainers/universal:2` image, we cannot run for example this sample on Codespaces: **[search application](https://github.com/microsoft/generative-ai-for-beginners/blob/main/08-building-search-applications/dotnet/notebook-azure-openai.dib?rgh-link-date=2024-01-12T02%3A59%3A09Z)**
> 

To unblock, can you use a diff base image and add the notebook features you need?

timheuer:
Related to #919

samruddhikhandale:
Update: Engaged the oryx team with https://github.com/microsoft/Oryx/pull/2272 which is needed to fix the oryx version compatibility issue. Currently, they are resolving issues with their own Oryx build pipelines. Once that is fixed and validations pass on ^ the PR, they are ready to merge it in.

src/universal/.devcontainer/devcontainer.json
```diff
@@ -9,10 +9,10 @@
             "userUid": "1000",
             "userGid": "1000"
         },
-        "ghcr.io/devcontainers/features/dotnet:1": {
-            "version": "7.0.306",
-            "installUsingApt": "false",
-            "additionalVersions": "6"
+        "ghcr.io/devcontainers/features/dotnet:2": {
+            "version": "8.0.x",
+            "dotnetRuntimeVersions": "7.0.x, 6.0.x",
```
samruddhikhandale, line 14:
```suggestion
            "dotnetRuntimeVersions": "6.0",
```

For the universal image, we only add support for two of the latest LTS versions. Hence, can we update this?

src/universal/.devcontainer/devcontainer.json
```diff
@@ -10,8 +10,9 @@
             "userGid": "1000"
         },
         "ghcr.io/devcontainers/features/dotnet:2": {
-            "version": "8.0",
-            "additionalVersions": "6.0"
+            "version": "8.0.x",
+            "dotnetRuntimeVersions": "7.0.x, 6.0.x",
+            "aspNetCoreRuntimeVersions": "7.0, 6.0.x"
```
samruddhikhandale, line 15: Same comment here!

src/universal/.devcontainer/devcontainer.json
timheuer: One of the reasons here @samruddhikhandale is that the likely inclusion of folks adding the VS Code extensions for .NET (C#/C# Dev Kit) require .NET 7 currently.  Work is underway to switch to .NET8, but until then it is an added convenience to have the runtimes in this image to avoid an additional acquisition that would slow the experience for using the devcontainer for .NET a bit.  .NET 7 is still fully supported right now.

samruddhikhandale: Thanks for the details, appreciate it. 

The `universal` image shouldn't support three versions of a particular tool/framework, at the most it should only support two LTS versions.

> .NET 7 is still fully supported right now.

I agree, however, .NET 7 is STS and not LTS. For `universal` images we aim at supporting only LTS versions.

Given the above reasoning, for a while we can ignore the rule of having only LTS and support .Net 7, but we want to avoid adding a third version (the image is already huge in size, would like to avoid expanding it)

@timheuer would adding .Net 8 and .Net 7 to the image sound good to you?

timheuer: I'd feel good about that decision.

samruddhikhandale:
```suggestion
            "dotnetRuntimeVersions": "7.0",
```

src/universal/.devcontainer/devcontainer.json
seesharprun: @samruddhikhandale, I updated the configuration to only include .NET 8 and 7.

src/universal/.devcontainer/devcontainer.json
```diff
@@ -9,10 +9,10 @@
             "userUid": "1000",
```
samruddhikhandale, line 9: Once https://github.com/devcontainers/features/pull/825 is merged, 👇 will most likely need changes similar to [this](https://github.com/devcontainers/features/pull/825/files#diff-bbf727c05eae41bfd55ff32120a1f879e516f178612ff6f7de5357831c0a9a1fR29-R31)

https://github.com/devcontainers/images/blob/8a52f4f05b222280e62ed7ce7410c728e559c5f4/src/universal/.devcontainer/local-features/setup-user/install.sh#L62-L66

seesharprun: Made the change and am testing now

samruddhikhandale:
https://github.com/devcontainers/images/pull/855 is merged, and new Oryx Feature is released.

@seesharprun Can you help update src/universal/.devcontainer/devcontainer-lock.json ?

seesharprun:
> #855 is merged, and new Oryx Feature is released.
> 
> @seesharprun Can you help update src/universal/.devcontainer/devcontainer-lock.json ?

Done:

https://github.com/seesharprun/devcontainers-images/blob/universal-dotnet-8/src/universal/.devcontainer/devcontainer-lock.json#L63-L67

samruddhikhandale:
@seesharprun Let me know if you need help with fixing the failing checks!

seesharprun:
> @seesharprun Let me know if you need help with fixing the failing checks!

I can't seem to figure it out. I would appreciate the assistance. I have selected the option to allow maintainers to edit.

image

samruddhikhandale (approved):
Looks good now, merged in changes which makes the test happy ⚡ 

I'll release dev tags today, and prod tags on Monday.

PR #860, opened 2023-11-19
Add support for Ruby extension

ondfisk:
This pull request adds support for the "shopify.ruby-extensions-pack" extension in the devcontainer.json file. This extension provides additional functionality for Ruby development in Visual Studio Code. By including this extension in the devcontainer.json file, it will be automatically installed when the container is created, making it easier for developers to work with Ruby code in the development environment. This change improves the development experience for Ruby developers using this repository.

### DIFF
--- a/src/ruby/.devcontainer/devcontainer.json
+++ b/src/ruby/.devcontainer/devcontainer.json
@@ -24,7 +24,7 @@
         "vscode": {
             // Add the IDs of extensions you want installed when the container is created.
             "extensions": [
-                "rebornix.Ruby"
+                "shopify.ruby-extensions-pack"
             ]
         }
     },


src/ruby/.devcontainer/devcontainer.json
```diff
@@ -24,7 +24,7 @@
         "vscode": {
             // Add the IDs of extensions you want installed when the container is created.
             "extensions": [
-                "rebornix.Ruby"
+                "shopify.ruby-extensions-pack"
```
samruddhikhandale, line 27: @bamurtaugh @craiglpeters Any thoughts on adding this extension?

> This is an extension pack that bundles together a set of extensions useful for Ruby development. It might include extensions for Ruby language support, code formatting, linting, debugging, and other tools that can help with Ruby development.

src/ruby/.devcontainer/devcontainer.json
```diff
@@ -24,7 +24,7 @@
         "vscode": {
             // Add the IDs of extensions you want installed when the container is created.
             "extensions": [
-                "rebornix.Ruby"
```
samruddhikhandale, line 27: Removing this makes sense to me!

src/ruby/.devcontainer/devcontainer.json
samruddhikhandale: If we decide to add it, then can we add this to the Ruby Feature instead? 

https://github.com/devcontainers/features/blob/main/src/ruby/devcontainer-feature.json#L21-L27

samruddhikhandale (commented):
Thanks for the contribution, left some comments.

src/ruby/.devcontainer/devcontainer.json
bamurtaugh: Thanks for opening this PR @ondfisk and for looping me in @samruddhikhandale!

I love the idea of ensuring folks have the right set of extensions, though it's certainly a trickly problem to tackle. Since I'm not sure we've seen other requests to add this extension pack, my first instinct is we leave this PR open for further community feedback. 

@ondfisk it could also be super helpful to share it in the community Slack channel to see if other Ruby dev container users would generally want this change: https://aka.ms/dev-container-community.

src/ruby/.devcontainer/devcontainer.json
bamurtaugh: Agree!

src/ruby/.devcontainer/devcontainer.json
eljog: We have a new issue requesting this:
https://github.com/devcontainers/images/issues/870

bamurtaugh: Thanks @eljog. Since we now have a couple requests for this, and I don't think I've seen any concerns against it, I'm fine with merging this change. What do you think @eljog?

bamurtaugh: Or I suppose the other consideration as @samruddhikhandale mentioned was adding it to the Feature instead of the image, which makes sense to me.

eljog: The feature already got the extension updated.
https://github.com/devcontainers/features/blob/3ea4d6bbd7864bcf7b5a91fdeeb66e4f5a6f46c0/src/ruby/devcontainer-feature.json#L21-L27

So, can we remove the extension from the image in this PR?
@ondfisk also please verify - the feature is installing `"shopify.ruby-lsp` extension, but your proposal is to add `shopify.ruby-extensions-pack`
I assume the pack is coming with additional extensions. Any preference over one or the other?

eljog: @joshaber in relation to your PR, what you do think about the above suggestion?

joshaber: I'm less certain about `shopify.ruby-extensions-pack` because I haven't used it, but we should definitely (at minimum) remove `rebornix.Ruby` since it's deprecated and since it sounds like the feature is installing the replacement extension.

joshaber: I updated https://github.com/devcontainers/images/pull/860 to just remove `rebornix.Ruby` since that seems like an obvious first step, regardless of where we land on `shopify.ruby-extensions-pack` 👍

eljog: We have got the asks for this PR addressed as mentioned at https://github.com/devcontainers/images/issues/870#issuecomment-1863805047 
I think we can close this PR in favor of the changes in the image and the ruby feature combined

AlvaroRausell:
Closing as stale

PR #864, opened 2023-11-28
Support PHP 8.3

samruddhikhandale:
Ref: https://github.com/devcontainers/images/issues/828

Adds support for PHP 8.3 for the `php` and `universal` images

PR #865, opened 2023-11-29
Add Java 21 to universal container

timward60:
Adds Java 21 as a feature to the universal container.

#770

### DIFF
--- a/src/universal/.devcontainer/devcontainer.json
+++ b/src/universal/.devcontainer/devcontainer.json
@@ -44,7 +44,7 @@
         },
         "ghcr.io/devcontainers/features/java:1": {
             "version": "21",
-            "additionalVersions": "11,17",
+            "additionalVersions": "17",
             "installGradle": "true",
             "installMaven": "true"
         },
--- a/src/universal/.devcontainer/devcontainer.json
+++ b/src/universal/.devcontainer/devcontainer.json
@@ -43,8 +43,8 @@
             "additionalVersions": "3.1"
         },
         "ghcr.io/devcontainers/features/java:1": {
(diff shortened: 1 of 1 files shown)


src/universal/.devcontainer/devcontainer.json
```diff
@@ -43,8 +43,8 @@
             "additionalVersions": "3.1"
         },
         "ghcr.io/devcontainers/features/java:1": {
-            "version": "17",
-            "additionalVersions": "11",
+            "version": "21",
+            "additionalVersions": "11,17",
```
timward60, line 47: **todo**: I cannot find an example of the expected format for a list here, looking for where this is defined.

samruddhikhandale, line 47:
```suggestion
            "additionalVersions": "17",
```

For the universal image, we only support two latest LTS versions, so 21 and 17 makes sense. Can we make this change? Thanks!

samruddhikhandale: `additionalVersions` is an experimental Feature option, but looks like this works.

samruddhikhandale (commented):
Thanks for contributing this PR, left one minor comment.

timward60:
@samruddhikhandale Thank you for the review. Applied your suggestion.

samruddhikhandale (approved):
Thank you! 🎉 

The test failure https://github.com/devcontainers/images/actions/runs/7065223046/job/19235282277?pr=865#step:3:41740 is similar to https://github.com/devcontainers/images/pull/864#issuecomment-1834322144.

PR #868, opened 2023-12-01
Tool for generating image release changes

eljog:
https://github.com/devcontainers/internal/issues/68

This pull request introduces a new script, `prepare-release.sh`, in the `build` directory. The script is designed to automate the process of preparing a release of the dev containers. It includes functionality to bump the version of the `manifest.json` file and run the `devcontainer upgrade` command. The script operates on either all images (useful for monthly release) or only those that have been modified since the last release commit (useful for ad-hoc release)

**The release PR https://github.com/devcontainers/images/pull/869 is created using this tool**

* <a href="diffhunk://#diff-f80c39bf43ffeebef6eee854c376f3f5fc899182c6154b9cb4e38275fda11249R1-R79">`build/prepare-release.sh`</a>: A new shell script that prepares a release of the dev containers. The script includes functions to get modified images, get all images, bump the version of the `manifest.json` file, and release an image. It also includes two types of releases: adhoc and monthly. The adhoc release operates on images modified since the last release commit, while the monthly release operates on all images. The script accepts a commit hash as input. If no commit hash is provided, it defaults to an monthly release.

Example adhoc release:
 `./build/prepare-release.sh 1c6f558dc86aafd7749074ec44e238f331303517`
Example monthly release:
 `./build/prepare-release.sh`

Note: It only bumps patch version. If you need to bump minor/major versions, that needs to be done manually.

image

### DIFF
--- a/build/prepare-release.sh
+++ b/build/prepare-release.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+
+# This script is used to prepare a release of the dev containers.
+# It will bump the version of the manifest.json file and run the devcontainer upgrade command.
+# It will only run on the images that have been modified since the last release commit that is passed in as the first argument.
+# If the second argument is "monthly", it will run on all images.
+# Example adhoc release: ./build/prepare-release.sh 1c6f558dc86aafd7749074ec44e238f331303517
+# Example monthly release: ./build/prepare-release.sh 1c6f558dc86aafd7749074ec44e238f331303517 monthly
+
+
+SCRIPT_SOURCE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+SRC_DIR=$(readlink -m $SCRIPT_SOURCE_DIR/../src)
+MANIFEST_FILE="manifest.json"
+COMMIT_HASH=$1
+RELEASE_TYPE=$2
+
+get_modified_images() {
+	git diff --name-only --diff-filter=ACMRTUB ${COMMIT_HASH} HEAD ${SRC_DIR} | while read file; do
+		if [ ! -z $file ]; then
+			commitMessage=$(git log -1 $file)
+			# omit auto commits from bot
+			if [[ $commitMessage != *"Dev containers Bot"* ]]; then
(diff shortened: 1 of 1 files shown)


build/prepare-release.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+
+# This script is used to prepare a release of the dev containers.
+# It will bump the version of the manifest.json file and run the devcontainer upgrade command.
+# It will only run on the images that have been modified since the last release commit that is passed in as the first argument.
+# If the second argument is "monthly", it will run on all images.
+# Example adhoc release: ./build/prepare-release.sh 1c6f558dc86aafd7749074ec44e238f331303517
+# Example monthly release: ./build/prepare-release.sh 1c6f558dc86aafd7749074ec44e238f331303517 monthly
```
samruddhikhandale, line 8: Curious, why do we need commit # for monthly release?

build/prepare-release.sh
```diff
@@ -0,0 +1,79 @@
...
+			fi
+		fi
+	done | sort | uniq
+}
+
+get_all_images() {
+	find $SRC_DIR -maxdepth 1 -type d | tail -n +2 | sort | uniq
+}
+
+bump_version() {
```
samruddhikhandale, line 36: This will always bump a `patch` version, right? So in any case, we'd need to look at commit history and decide if it needs `major` or `minor` bump?

samruddhikhandale (approved):
This is wonderful, thanks for this enhancement 🎉 

Looks great, left some minor comments.

build/prepare-release.sh
eljog: You are right.. it is not needed
Let me modify

build/prepare-release.sh
eljog: Correct
For now, anything other than patch needs to be manually updated after generating

build/prepare-release.sh
eljog: addressed in https://github.com/devcontainers/images/commit/6b1d82a57a306659f43b188aeca144125d03e596

joshspicer (approved):
Re approving

PR #831, opened 2023-11-03
[universal] Bump Python SDK versions

alexander-smolyakov:
**Devcontainer name**: 

- universal

**Description**:

With the recent [release of Oryx](https://github.com/microsoft/Oryx/releases/tag/20231102.1), newer versions of Python are [now supported](https://github.com/microsoft/Oryx/pull/2237). Thus, we can bump up the Python SDK versions we ship with the universal devcontainer and remove some security patches.

**Changelog**:

_Bumped Python SDK versions_:

- `v3.10`: `3.10.8` -> `3.10.13`;
- `v3.9`: `3.9.16` -> `3.9.18`;

_`patch-conda` feature_: 

- Removed outdated patches;
- Added `pip show` command to output additional information about patched packages into build logs;

_`patch-python` feature_: 

- Removed outdated patches;
-  Added `pip show` command to output additional information about patched packages into build logs;

_`test-utils.sh`_:

- Reworked the `checkPythonPackageVersion` function to make it generic for all Python packages. Now, the function uses the `importlib.metadata` [library](https://docs.python.org/3/library/importlib.metadata.html#distribution-versions) to get the package version instead of relying on the `__version__` attribute.

**Checklist**:

- [ ] Checked that applied changes work as expected

PR #876, opened 2023-12-06
[universal] Update `setuptools` for Python 3.10 due to GHSA-r9hx-vwmv-q579

alexander-smolyakov:
**Devcontainer name**: 

- universal

**Description**:

Recently, a test for verifying the `setuptools` package version started to fail due to the `setuptools` version downgrade. The original patch was removed in the following PR because the Python v3.10.13 version was shipped with `setuptools v68.2.2`:

- #831

It seems that the Python v3.10.13 distribution was changed and the `setuptools` package was downgraded for some reason.

*Changelog*:

- Updated `patch-python` feature to restore patch for `setuptools` package (The version is set to `68.2.2` to align it with [v2.6.0](https://github.com/devcontainers/images/blob/main/src/universal/history/2.6.0.md));

**Checklist**:


eljog (approved):
LGTM

Issue #870, opened 2023-12-05
The  `rebornix.Ruby` extension is marked as deprecated


jcansdale:
The [rebornix.Ruby](https://marketplace.visualstudio.com/items?itemName=rebornix.Ruby) extension is marked as deprecated. It suggests moving to the [shopify.ruby-lsp](https://marketplace.visualstudio.com/items?itemName=shopify.ruby-lsp) extension instead.

![Image](https://github.com/devcontainers/images/assets/11719160/e61d81be-30e8-4dfb-b823-17de20aa1abf)

Should we update the `devcontainer.json` here?
https://github.com/devcontainers/images/blob/main/src/ruby/.devcontainer/devcontainer.json#L27

eljog:
Apparently we already have a PR out for this https://github.com/devcontainers/images/pull/860
We have been waiting to learn more about the preferences of the community
https://github.com/devcontainers/images/pull/860/files#r1406569477
>it's certainly a trickly problem to tackle as everyone has their own set of preferences. Since I'm not sure we've seen other requests to add this extension pack, my first instinct is we leave this PR open for further community feedback.

eljog:
#894 has removed the deprecated extension for devcontainer, and the ruby feature [here](https://github.com/devcontainers/features/blob/3ea4d6bbd7864bcf7b5a91fdeeb66e4f5a6f46c0/src/ruby/devcontainer-feature.json#L21-L27) is already installing the suggested [shopify.ruby-lsp](https://marketplace.visualstudio.com/items?itemName=shopify.ruby-lsp) extension.
🎉

PR #894, opened 2023-12-18
Remove deprecated Ruby extension

joshaber:
Helps https://github.com/devcontainers/images/issues/870

Does what it says on the tin. `rebornix.Ruby` has been deprecated and the Ruby feature is already installing the replacement: https://github.com/devcontainers/features/blob/3ea4d6bbd7864bcf7b5a91fdeeb66e4f5a6f46c0/src/ruby/devcontainer-feature.json#L21-L27

### DIFF
--- a/src/ruby/.devcontainer/devcontainer.json
+++ b/src/ruby/.devcontainer/devcontainer.json
@@ -24,7 +24,7 @@
         "vscode": {
             // Add the IDs of extensions you want installed when the container is created.
             "extensions": [
-                "rebornix.Ruby"
+                "Shopify.ruby-lsp"
             ]
         }
     },
--- a/src/ruby/.devcontainer/devcontainer.json
+++ b/src/ruby/.devcontainer/devcontainer.json
@@ -23,9 +23,7 @@
         // Configure properties specific to VS Code.
         "vscode": {
             // Add the IDs of extensions you want installed when the container is created.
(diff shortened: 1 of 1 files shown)


src/ruby/.devcontainer/devcontainer.json
```diff
@@ -24,7 +24,7 @@
         "vscode": {
             // Add the IDs of extensions you want installed when the container is created.
             "extensions": [
-                "rebornix.Ruby"
+                "Shopify.ruby-lsp"
```
eljog, line 27: I have also mentioned you on a related discussion here https://github.com/devcontainers/images/pull/860#discussion_r1423276351

What do you think about that approach?

joshaber: I updated this to just remove `rebornix.Ruby` since that seems like a good first step, regardless of where the conversation in https://github.com/devcontainers/images/pull/860 ends up.

eljog: Sounds good
Now there there is are no `customizations`, can we remove/comment-out the entire `customizations` object

joshaber: Done! 💥

eljog (approved):
⚡ Thank you for contributing to the devcontainers community

PR #920, opened 2024-01-12
Adds .NET8 to Universal

timheuer:
Universal still using .NET 7 SDK and didn't get update with the .NET8 release.  This adds that support while also adding .NET7 additional runtime support which is currently used by some extensions that are installed into devcontainers.


eljog:
Thank you for contributing.
But, there exists a PR proposing the same https://github.com/devcontainers/images/pull/855 , but is blocked due to an `Oryx` concern. Please see the comments on the other PR for details.

timheuer:
Excellent, I'll close this one and comment on the other in-progress!

PR #956, opened 2024-02-08
Fix image release workflows and explicitly add permission blocks

samruddhikhandale:
Fixes workflow runs - https://github.com/devcontainers/images/actions/runs/7834841668

image


.github/workflows/push.yml
```diff
@@ -72,6 +72,9 @@ jobs:
     name: Trigger info extraction
     needs: [build-and-push]
     uses: devcontainers/images/.github/workflows/version-history.yml@main
+    permissions:
```
eljog, line 75: Did anything change with `version-history.yml`, for it to require these permissions?
Or are the permissions reset somewhere else?

eljog (approved):

.github/workflows/push.yml
samruddhikhandale: Discussed async

PR #965, opened 2024-02-14
[MIniconda] - Cryptography - Patch security vulnerability - GHSA-3ww4-gg4f-jr7f

gauravsaini04:
**Dev container name**:
 
 * Miniconda
 
 **Description**:
 
 This PR patches the following vulnerability:
 
 * [GHSA-3ww4-gg4f-jr7f](https://github.com/advisories/GHSA-3ww4-gg4f-jr7f) - related to the `cryptography` package;
 
 This vulnerability comes from the coninuumio/miniconda3 image used upstream for the Miniconda devcontainer.
 
 _Changelog_:
 
 * Updated Dockerfile
   * Upgraded version for patched conda package;
     * `cryptography` - _minimum package version is set to `42.0.2`_;
   * Upgraded version for patched python package;
     * `pyopenssl` - _minimum package version is set to `24.0.0`_  ( _for compatibility with `cryptography v42.0.2`_ );
	 
 * Updated tests to verify the following: 
    * `cryptography` minimum version (Minimum package version set to `42.0.2` which fixes [GHSA-3ww4-gg4f-jr7f](https://github.com/advisories/GHSA-3ww4-gg4f-jr7f));
    * `pyopenssl` minimum version (Minimum package version set to `24.0.0` which is compatible with `cryptography v42.0.2`;
 
 **Checklist**:
 

src/miniconda/.devcontainer/Dockerfile
```diff
@@ -7,8 +7,8 @@ FROM continuumio/miniconda3 as upstream
     # <package_name> = <version>
 
 RUN conda install \
-    # https://github.com/advisories/GHSA-jfhm-5ghh-2f97
-    cryptography==41.0.7
+    # https://github.com/advisories/GHSA-3ww4-gg4f-jr7f
+    cryptography==42.0.2
```
samruddhikhandale, line 11: Question: The patched version is `42.0.0`, so why was it updated to `42.0.2`?

samruddhikhandale (changes requested):

src/miniconda/test-project/test.sh
```diff
@@ -18,11 +18,11 @@ check "gitconfig-contains-name" sh -c "cat /etc/gitconfig | grep 'name = devcont
 
 check "usr-local-etc-config-does-not-exist" test ! -f "/usr/local/etc/gitconfig"
 
-checkPythonPackageVersion "cryptography" "41.0.7"
+checkPythonPackageVersion "cryptography" "42.0.2"
 checkPythonPackageVersion "setuptools" "65.5.1"
 checkPythonPackageVersion "wheel" "0.38.1"
```
samruddhikhandale, line 24: There are failing tests, can you take a look? https://github.com/devcontainers/images/actions/runs/7899131830/job/21557930111?pr=965#step:3:2888

src/miniconda/.devcontainer/Dockerfile
gauravsaini04: Because it is preferred to get the latest version even if higher than the required version from the anaconda channel and I could only find this version after 41.0.7 in anaconda channel for cryptography and not exactly 42.0.0. ( alex had mentioned like that to me )

src/miniconda/test-project/test.sh
gauravsaini04: Have removed the test for `checkCondaPackageVersion "pyopenssl" "23.2.0"` as no pyopenssl package was found inside the dev container when checked with commands like `pip show pyopenssl` or `conda list | grep -i pyopenssl`

samruddhikhandale: Installing new version of `crypotography` removed `pyopenssl`, see https://github.com/devcontainers/images/actions/runs/7899131830/job/21557930111#step:3:719

That doesn't sound good and could be a breaking change for users who were previously expecting `pyopenssl`

Can you install a compatible version of `pyopenssl` which works with `crypotography` ? See https://github.com/devcontainers/images/pull/634 for reference

See https://github.com/devcontainers/images/issues/486 for a past problem due to version incompatibility.

src/miniconda/.devcontainer/Dockerfile
samruddhikhandale: > Because it is preferred to get the latest version

As we are patching, it is not preferred in our case to always install `latest`. We want to get to a point where the upstream image updates it and we can remove this patch. 

> I could only find this version after 41.0.7 in anaconda channel for cryptography and not exactly 42.0.0

This explanation sounds reasonable

samruddhikhandale (changes requested):

src/miniconda/test-project/test.sh
gauravsaini04: Have installed `pyopenssl v24.0.0` as compatible version with `cryptography v42.0.2` (as mentioned in https://pypi.org/project/pyOpenSSL/)

src/miniconda/.devcontainer/Dockerfile
```diff
@@ -7,8 +7,12 @@ FROM continuumio/miniconda3 as upstream
...
+
+RUN python3 -m pip install --upgrade \
+    # installed for compatibility with cryptography v42.0.2
+    pyopenssl==24.0.0
```
samruddhikhandale, line 15: Curious, why don't we install it with `conda`? Similar to https://github.com/devcontainers/images/pull/634/files

samruddhikhandale (changes requested):

src/miniconda/.devcontainer/Dockerfile
gauravsaini04: saw compatible versions for `cryptography v42.0.2` on https://pypi.org/project/pyOpenSSL/#section-3
and unavailability of `pyopenssl` required version on https://anaconda.org/search?q=pyopenssl
so, installed with pip

samruddhikhandale: I see, can you add another test  for `pip` and make sure pip installs continue to work? 

similar to https://github.com/devcontainers/images/blob/e7e4b123bc46bd2f4a265ceb30b8c617c313ec03/src/miniconda/test-project/test.sh#L32-L34

samruddhikhandale (changes requested):

src/miniconda/.devcontainer/Dockerfile
gauravsaini04: Added tests to check that pip install works in correct manner.

samruddhikhandale (approved):
Looks good!

PR #980, opened 2024-02-29
fix indent on line 15

elion-abdyli:
Hi, 

I've never done a PR before except for school projects.
I was reading the file and I think I noticed an indentation typo.
If this gets accepted it will be my first ever contribution to a real life project !

Thanks !


elion-abdyli (commented):
Unindented once. Shouldn't affect function, only whitespace was removed.

samruddhikhandale (approved):
🎉

PR #1002, opened 2024-03-20
Adding inotify-tools to universal image

osortega:
This PR adds `inotify-tools` to the list of packages to be installed for the universal image.

### DIFF
--- a/src/universal/.devcontainer/Dockerfile
+++ b/src/universal/.devcontainer/Dockerfile
@@ -55,6 +55,7 @@ RUN apt-get update \
         uuid-dev \
         curl \
         gettext \
+        inotify-tools \
     && rm -rf /var/lib/apt/lists/* \
     # This is the folder containing 'links' to benv and build script generator
     && apt-get update \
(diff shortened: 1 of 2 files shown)


joshspicer (approved):
LGTM

src/universal/.devcontainer/Dockerfile
```diff
@@ -55,6 +55,7 @@ RUN apt-get update \
         uuid-dev \
         curl \
         gettext \
+        inotify-tools \
```
samruddhikhandale, line 58: Can you add a test?
 
You can add this package to the list [here](https://github.com/devcontainers/images/blob/c4220a0e94143536bcbe9490662e8e47ba2fb43f/src/universal/test-project/test-utils.sh#L133) for testing. (This way we can validate if things change in future (like soon we might change the base image from focal to bookworm, would be good to have a validation in place.)

samruddhikhandale (changes requested):
Left one minor comment, otherwise it looks great!

samruddhikhandale (approved):
Thanks! I'll merge this once tests are 🟢

PR #1008, opened 2024-03-29
Spelling correction in universal images README.md

adamhu714:
"languages lke Python" is changed to "languages like Python".


samruddhikhandale (approved):
Thanks!

Issue #1016, opened 2024-04-02
Support Ubuntu 24.04 scheduled for April 25th


samruddhikhandale:
- [x] Open a PR 
- [x] Release image
- [x] Update [templates](https://github.com/devcontainers/templates) (if necessary)
- [x] Update https://github.com/devcontainers/images/issues/90
- [x] Update [Microsoft Artifact Registry](https://mcr.microsoft.com/en-us/catalog?search=dev%20container) docs

samruddhikhandale:
Opened https://github.com/devcontainers/images/pull/1036 which would be released this week.

samruddhikhandale:
Released [mcr.microsoft.com/devcontainers/base:noble](https://github.com/devcontainers/images/tree/main/src/base-ubuntu) 🎉

samruddhikhandale:
Closing as completed 🎉

PR #1036, opened 2024-04-25
Support Ubuntu 24.04 (noble)

samruddhikhandale:
Ref: https://github.com/devcontainers/images/issues/1016 and https://wiki.ubuntu.com/Releases

- Supports the latest LTS version of Ubuntu
- Once released, `mcr.microsoft.com/devcontainers/base:ubuntu` will now point to `noble` instead of `jammy` as the `ubuntu` tag is meant to represent the latest LTS version

### DIFF
--- a/src/base-ubuntu/manifest.json
+++ b/src/base-ubuntu/manifest.json
@@ -1,6 +1,7 @@
 {
-	"version": "1.0.23",
+	"version": "1.1.0",
 	"variants": [
+		"noble",
 		"jammy",
 		"focal"
 	],
@@ -8,6 +9,10 @@
 		"latest": false,
 		"rootDistro": "debian",
 		"architectures": {
+			"noble": [
+				"linux/amd64",
+				"linux/arm64"
+			],
 			"jammy": [
 				"linux/amd64",
 				"linux/arm64"
@@ -20,10 +25,14 @@
 			"base:${VERSION}-${VARIANT}"
 		],
 		"variantTags": {
+			"noble": [
+				"base:${VERSION}-ubuntu-24.04",
+				"base:${VERSION}-ubuntu24.04",
+				"base:${VERSION}-ubuntu"
+			],
 			"jammy": [
 				"base:${VERSION}-ubuntu-22.04",
-				"base:${VERSION}-ubuntu22.04",
-				"base:${VERSION}-ubuntu"
+				"base:${VERSION}-ubuntu22.04"
 			],
 			"focal": [
 				"base:${VERSION}-ubuntu-20.04",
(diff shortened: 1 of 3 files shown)


src/base-ubuntu/manifest.json
```diff
@@ -20,10 +25,14 @@
 			"base:${VERSION}-${VARIANT}"
 		],
 		"variantTags": {
+			"noble": [
+				"base:${VERSION}-ubuntu-24.04",
+				"base:${VERSION}-ubuntu24.04",
+				"base:${VERSION}-ubuntu"
```
samruddhikhandale, line 31: > Ubuntu 24.04 LTS will be supported for 5 years until June 2029. If you need Long Term Support, we recommend you use [Ubuntu 22.04 LTS 728](https://wiki.ubuntu.com/JammyJellyfish/ReleaseNotes/) until 24.04.1 is released (which is scheduled for the 15th of August).

From https://discourse.ubuntu.com/t/noble-numbat-release-notes/39890#support-lifespan-3 , I wonder if we should hold off from making `noble` distro tagged with `ubuntu`

samruddhikhandale: @joshspicer Any thoughts?

src/base-ubuntu/.devcontainer/devcontainer.json
```diff
@@ -7,8 +7,6 @@
         "ghcr.io/devcontainers/features/common-utils:2": {
             "installZsh": "true",
             "username": "vscode",
-            "userUid": "1000",
```
joshspicer, line 10: I suppose we don't have a way to programatically validate that older versions of ubuntu still have uid 1000?  

I believe you validated it manually, so I think that should be good (we understand the behavior of `common-utils` and know it will add the uid at 1000 or the next available uid).

joshspicer (approved):

src/base-ubuntu/.devcontainer/devcontainer.json
samruddhikhandale: > I suppose we don't have a way to programatically validate that older versions of ubuntu still have uid 1000?

We don't :| 

> I believe you validated it manually, so I think that should be good (we understand the behavior of common-utils and know it will add the uid at 1000 or the next available uid).

Yes 👍

felipecrs: IMO the final user should be kept as UID 1000, at least for the base images like Ubuntu.

The ubuntu user that comes with 24.04 doesn't seem used by anything. I guess it was only created by convenience.

An easy way to fix this would be to delete the user prior to let common-utils run.

felipecrs: Like I did in https://github.com/felipecrs/docker-images/pull/144

PR #1073, opened 2024-05-22
[Anaconda] - remove pinned packages as availabe versions from upstream greater than pinned

gauravsaini04:
**Dev container name**:
 
 * Anaconda
 
 **Description**:
 
 This PR removes a number of pinned packages as available packages from upstream have greater or equal versions for them.
 
 _Changelog_:
 
 * Updated `Dockerfile`
 * Updated tests
 
 **Checklist**:
 

src/anaconda/test-project/test.sh
```diff
@@ -30,25 +30,18 @@ check "gitconfig-contains-name" sh -c "cat /etc/gitconfig | grep 'name = devcont
 check "usr-local-etc-config-does-not-exist" test ! -f "/usr/local/etc/gitconfig"
 
 checkPythonPackageVersion "joblib" "1.2.0"
-checkPythonPackageVersion "cookiecutter" "2.1.1"
```
samruddhikhandale, line 33: Can we keep the tests around? This way we can test if the removal is working, also would ensure that we don't install any previous version in future.

samruddhikhandale (changes requested):

src/anaconda/test-project/test.sh
gauravsaini04: sure reverted the test file changes..

samruddhikhandale (approved):
Nice 🧹

PR #1074, opened 2024-05-22
[Miniconda] - remove pinned packages - as higher versions availabe from upstream image

gauravsaini04:
**Dev container name**:
 
 * Miniconda
 
 **Description**:
 
 This PR removes a couple of pinned packages as similar available packages from upstream have greater version for them.
 
 _Changelog_:
 
 * Updated `Dockerfile`
 * Updated tests
 
 **Checklist**:
 
### DIFF
--- a/src/miniconda/test-project/test.sh
+++ b/src/miniconda/test-project/test.sh
@@ -23,7 +23,6 @@ checkPythonPackageVersion "setuptools" "65.5.1"
 checkPythonPackageVersion "wheel" "0.38.1"
 
 checkCondaPackageVersion "cryptography" "42.0.4"
-checkCondaPackageVersion "pyopenssl" "24.0.0"
 checkCondaPackageVersion "setuptools" "65.5.1"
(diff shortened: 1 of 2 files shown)


src/miniconda/test-project/test.sh
```diff
@@ -18,12 +18,9 @@ check "gitconfig-contains-name" sh -c "cat /etc/gitconfig | grep 'name = devcont
 
 check "usr-local-etc-config-does-not-exist" test ! -f "/usr/local/etc/gitconfig"
 
-checkPythonPackageVersion "cryptography" "42.0.4"
```
samruddhikhandale, line 21: Can we keep the tests around? This way we can test if the removal is working, also would ensure that we don't install any previous version in future.

samruddhikhandale (changes requested):

src/miniconda/test-project/test.sh
gauravsaini04: done

samruddhikhandale (approved):
🧹

Issue #989, opened 2024-03-06
Library search path related issue in latest miniconda env


bpkroth:
As described in https://github.com/microsoft/MLOS/pull/699, there appears to be a problem with the latest miniconda devcontainer image.  Somewhere in the stack additional environments added to it end up searching in the base debian image `/usr/lib` path for `libc.so` instead of those installed with `conda` (e.g., `/opt/conda/lib` or `/opt/conda/envs/$env_name/lib`).

Rolling back to 0.203.6-3 fixes the issue as does explicitly setting `LD_LIBRARY_PATH=/opt/conda/lib`.

samruddhikhandale:
Thanks for reporting! @gauravsaini04 can you prioritize investigating the issue? thanks!

bpkroth:
FYI, I made a smaller reproduction here:
https://github.com/bpkroth/miniconda-devcontainer-test

I think it might actually have something to do with pandas being installed from conda and pyarrow being installed from pip.

In the original example this was unintentional, but here I did it explicitly.

Anyways, still unclear why the base devcontainer change should affect that.

gauravsaini04:
> Thanks for reporting! [@gauravsaini04](https://github.com/gauravsaini04) can you prioritize investigating the issue? thanks!

sure @samruddhikhandale, will investigate it

samruddhikhandale:
Thanks @bpkroth for providing us more information. The repro repo is pretty helpful, `conda env update -n base -v -f environment.yml` command definitely fails with the latest image

> I think it might actually have something to do with pandas being installed from conda and pyarrow being installed from pip.

The only change that went into latest version of the miniconda image is https://github.com/devcontainers/images/pull/976  which did something similar to ^ ; where `cryptography` installation was moved from conda to pip. I am not confident if that's the root cause, however., @gauravsaini04  I can think of few ways to debug the root cause - 

1. Can you use https://github.com/bpkroth/miniconda-devcontainer-test for repro, and instead of using the base image, can you build the miniconda image with reverted https://github.com/devcontainers/images/pull/976 changes?
2. Can you check if there was a newer release of continuumio/miniconda3 which could have caused this?

gauravsaini04:
@prathameshzarkar9, had been working on this issue.
He has been testing the issue with the following scenarios, as suggested by you : 
Scenario #1 : Test with the following `environment.yml` in the test project by @bpkroth,
```
channels:
  - defaults
dependencies:
  - python
  - pandas
  - pytest
[...]

PR #1075, opened 2024-05-22
Library search path related issue fixed

prathameshzarkar9:
https://github.com/devcontainers/images/issues/989 Library search path related issue in latest miniconda env
**Dev container name**:
 * miniconda
 
**Description**:
 This PR fixes the issue of searching `libc.so` in the `/usr/lib` due to additional environments added in the stack
 _Changelog_:
 * Updated Dockerfile
   * added environment variable `LD_LIBRARY_PATH=/opt/conda/lib`

**Checklist**:
 * [x]   built new miniconda image and added used the same image for miniconda-devcontainer-test project to run.

### DIFF
--- a/src/miniconda/.devcontainer/Dockerfile
+++ b/src/miniconda/.devcontainer/Dockerfile
@@ -6,13 +6,11 @@ RUN conda install \
     # https://github.com/advisories/GHSA-jjg7-2v4v-x38h
     idna==3.7
 
-RUN conda install \
-    # https://github.com/advisories/GHSA-3ww4-gg4f-jr7f
-    cryptography==42.0.5
+RUN ENV LD_LIBRARY_PATH=/opt/conda/lib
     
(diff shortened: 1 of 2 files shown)


src/miniconda/.devcontainer/Dockerfile
```diff
@@ -6,6 +6,8 @@ RUN conda install \
     # https://github.com/advisories/GHSA-jjg7-2v4v-x38h
     idna==3.7
 
+ENV LD_LIBRARY_PATH=/opt/conda/lib
```
samruddhikhandale, line 9: Discussed this offline, let's research on the purpose and impact of setting `LD_LIBRARY_PATH`

samruddhikhandale:
Closing as no longer needed. See https://github.com/devcontainers/images/issues/989#issuecomment-2130521693

PR #1095, opened 2024-06-17
dotnet: improve dev-certs instructions

sliekens:
I updated the README for dotnet with a better way to manage HTTPS development certificates for ASP.NET Core.

Previously, it recommended a solution that did not work well in team or open source environments, as it required manual steps, and the solution carried a risk of unintended password disclosure.

I added a solution which does not require manual steps, nor does it have any risk of leaking passwords.

I did not remove the old instructions, but I did add a warning to avoid using it for shared projects.

PR #1103, opened 2024-06-20
dotnet: amend dev-cert script in README

sliekens:
Hello again,

This is a minor correction for the dev-cert instructions that I added earlier this week in #1095.

In the original instructions, `sudo` is used to generate a new dev-cert. As a result, the `vscode` user might be unable to use the dev-cert due to insufficient permissions. The dev-cert is owned by `root:root` and the file permissions are `-rw------` (600).

The solution is simply to run the `dotnet dev-certs https` command first as the `vscode` user, resulting in the file being created with correct ownership. The use of `sudo` is still needed to export the dev-cert to the `ca-certificates/` directory.


samruddhikhandale (approved):
Thanks!

PR #1108, opened 2024-06-28
[universal] - check toolings - monthly task

gauravsaini04:
**Dev container name**:
 
 * Universal
 
 **Description**:
 
 This PR updates the toolings versions for `Universal` dev container image to `long term supported` versions if available for the runtime, else uses the `latest` versions of those runtimes
 
 _Changelog_:
 
 * Updated `devcontainer.json` file
 
 **Checklist**:
 

khgi6565:
> **Dev container name**:
> 
> * Universal
> 
> **Description**:
> 
> This PR updates the toolings versions for `Universal` dev container image to `long term supported` versions if available for the runtime, else uses the `latest` versions of those runtimes
> 
> _Changelog_:
> 
> * Updated `devcontainer.json` file
> 
> **Checklist**:
> 
> * [x]    Checked that applied changes work as expected

src/universal/.devcontainer/devcontainer.json
```diff
@@ -23,8 +23,8 @@
         },
         "./local-features/nvs": "latest",
         "ghcr.io/devcontainers/features/python:1": {
-            "version": "3.10.13",
-            "additionalVersions": "3.9.18",
+            "version": "3.12.3",
```
samruddhikhandale, line 26: Oryx has compatibility issues with the version of Python installed, hence can we update the version based on the closest lts version which is supported by oryx?

Looking at https://github.com/devcontainers/images/actions/runs/9771395221/job/26974077297?pr=1108#step:3:38957 we'd need `3.12.1`

image

samruddhikhandale: That's the reason why we have tests to ensure the installed version is compatible with Oryx

src/universal/.devcontainer/local-features/patch-python/install.sh
```diff
@@ -38,13 +38,3 @@ update_package() {
     sudo_if "$PYTHON_PATH -m pip install --upgrade --no-cache-dir $PACKAGE==$VERSION"
     sudo_if "$PYTHON_PATH -m pip show --no-python-version-warning $PACKAGE"
 }
-
-# Temporary: Upgrade python packages due to security vulnerabilities
-# They are installed by the base image (python) which does not have the patch.
-
-# https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-40897
```
samruddhikhandale, line 45: Niceeee

samruddhikhandale (changes requested):

src/universal/.devcontainer/devcontainer.json
gauravsaini04: thanks for info, done

src/universal/.devcontainer/devcontainer.json
```diff
@@ -39,8 +39,8 @@
             "version": "latest"
         },
         "ghcr.io/devcontainers/features/ruby:1": {
-            "version": "3.2",
-            "additionalVersions": "3.1"
+            "version": "3.3.3",
```
samruddhikhandale, line 42:
```suggestion
            "version": "3.3.4",
```

Can we update to 3.3.4?  Hoping to not run into a similar issue as https://github.com/devcontainers/images/issues/1123

samruddhikhandale (changes requested):

samruddhikhandale (approved):

PR #1111, opened 2024-07-05
Update actions: Use token based authentication

samruddhikhandale:
Switch `az login` authentication from Service Principal mechanism to token based mechanism.

Ref: https://learn.microsoft.com/en-us/azure/container-registry/container-registry-repository-scoped-permissions

### DIFF
--- a/.github/workflows/push-again.yml
+++ b/.github/workflows/push-again.yml
@@ -16,6 +16,7 @@ jobs:
     name: Build and push images
     if: ${{ github.ref == 'refs/heads/main' }}
     runs-on: ubuntu-latest
+    environment: publishing
     steps:
     - name: Free more space
       id: free_space 
@@ -39,12 +40,6 @@ jobs:
       with:
         path: 'release'
         ref: ${{ github.event.inputs.release }}
-
-    - name: Azure CLI login
-      id: az_login
-      uses: azure/login@v1
-      with:
-        creds: ${{ secrets.AZ_ACR_CREDS }}
       
     - name: Build and push
       id: build_and_push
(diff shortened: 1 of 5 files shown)


.github/workflows/version-history.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     name: Update version history
     if: ${{ github.ref == 'refs/heads/main' || github.event.base_ref == 'refs/heads/main' }}
     runs-on: ubuntu-latest
-    environment: documentation
+    environment: publishing
```
chrmarti, line 32: Is the token associated with the `publishing` environment?

.github/workflows/push-again.yml
```diff
@@ -54,12 +49,14 @@ jobs:
         STUB_REGISTRY: ${{ secrets.STUB_REGISTRY }}
         STUB_REGISTRY_BASE_PATH: ${{ secrets.STUB_REGISTRY_BASE_PATH }}
         SECONDARY_REGISTRY_BASE_PATH: ${{ secrets.SECONDARY_REGISTRY_BASE_PATH }}
+        TOKEN_NAME: ${{ secrets.TOKEN_NAME }}
+        PASSWORD: ${{ secrets.PASSWORD }}
```
chrmarti, line 53: Does the token have an expiration date?

chrmarti (approved):
LGTM!

Are we using the ACR to let MCR pull from it or something similar? Would GHCR work instead?

.github/workflows/push-again.yml
samruddhikhandale: Yes, it's similar to authenticating with Service Principals (~one year)

Once we move to another tenant, we'd be able to use Managed Identities which doesn't need password rotations.

chrmarti:
> > Would GHCR work instead?
> 
> Won't we need to push our images to GHCR for that to work? 🤔 I don't think we'd do that as the images would break.

I guess we can't use GHCR to push to MCR? That would avoid the need for the ACR credentials and the GitHub workflows can easily be given access to their repository's namespace in GHCR.

But this would be longer term. I agree with the approach taken here. 👍

samruddhikhandale:
> > > Would GHCR work instead?
> > 
> > 
> > Won't we need to push our images to GHCR for that to work? 🤔 I don't think we'd do that as the images would break.
> 
> I guess we can't use GHCR to push to MCR? That would avoid the need for the ACR credentials and the GitHub workflows can easily be given access to their repository's namespace in GHCR.
> 
> But this would be longer term. I agree with the approach taken here. 👍

Ah, got it. When I get to researching the moving of tenants task, I can see if this is even possible. Thanks for the info here.

PR #1127, opened 2024-07-12
[cpp] - buster - EOL

gauravsaini04:
**Devcontainer name**

* CPP

**Changes**

* Changes to manifest.json for CPP dev container image's files for buster EOL

### DIFF
--- a/src/cpp/manifest.json
+++ b/src/cpp/manifest.json
@@ -3,7 +3,6 @@
 	"variants": [
 		"bookworm",
 		"bullseye",
-		"buster",
 		"noble",
 		"jammy",
 		"focal"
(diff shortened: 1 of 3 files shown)


src/cpp/manifest.json
```diff
@@ -3,7 +3,6 @@
 	"variants": [
 		"bookworm",
```
samruddhikhandale, line 4: Can we cleanup from README file as well?

samruddhikhandale, line 4: Can we cleanup https://github.com/devcontainers/images/blob/fea177cf129ebfe01c7b50263fc1a475f0bd3ea3/src/cpp/.devcontainer/base-scripts/install-vcpkg.sh#L14relevant code ?

samruddhikhandale (changes requested):

src/cpp/manifest.json
gauravsaini04: done

gauravsaini04: done

samruddhikhandale (approved):

PR #1164, opened 2024-08-16
[Universal] - install latest cpu torch

gauravsaini04:
**Devcontainer Name**

* Universal 

**Description**

* This PR attempts to install pytorch 2.4.0 i.e. the latest version with cpu only acceleration limiting it to not use nvidia-cuda and triton packages, thereby reducing the size of the image considerably.

**_Changelog_**

* Added pytorch 2.4.0 along with necessary packages for cpu accelerated latest version of torch package

**Checklist**

* [x] Checked that applied changes work as expected


src/universal/.devcontainer/local-features/machine-learning-packages/install.sh
```diff
@@ -44,7 +44,7 @@ if [[ "$(python --version)" != "" ]] && [[ "$(pip --version)" != "" ]]; then
     install_python_package "matplotlib"
     install_python_package "seaborn"
     install_python_package "scikit-learn"
-    install_python_package "torch==2.3.1" "-f https://download.pytorch.org/whl/cpu/torch_stable.html"
+    install_python_package "torch torchvision torchaudio" "--index-url https://download.pytorch.org/whl/cpu"
```
samruddhikhandale, line 47: Curious, why do we need to add installations for `torchvision` and `torchaudio` ?

gauravsaini04: actually, torch when reduced to cpu accelerated version only, so including `torch vision` and `torchaudio`, since they provide datasets, models, and transforms for computer vision and audio tasks etc. and therefore, Installing all three packages together can ofcourse avoid future potential compatibility issues that might arise later. I included them as we had reduced `torch` to `cpu accelerated` version only, so atleast by adding these alongside `torch` will provide some functionality here as well. But we could ofcourse only install torch and not these additional packages.

samruddhikhandale: Thanks for the details. Since `torchvision` and `torchaudio` enhance the functionality of `torch` specifically for domains like images and audio, let's hold off on adding those until we hear from users. I believe the CPU-only package won't be very helpful for these operations, so the user will likely end up installing the GPU variant. Let's skip it for now. Thanks!

samruddhikhandale, line 47:
```suggestion
    install_python_package "torch" "--index-url https://download.pytorch.org/whl/cpu"
```

samruddhikhandale (approved):

PR #1176, opened 2024-08-27
Create SECURITY.md

goxijana3:


bamurtaugh:
Thanks for starting this! If you have specific content you'd like covered in a SECURITY.md, please let us know, otherwise looks like this is a general template (so closing for now).

PR #1224, opened 2024-10-30
fix: idea default dir can't use git

Mr-Ao-Dragon:


Mr-Ao-Dragon:

chrmarti:
This seems specific to where you mount or clone your git repositories with IDEA which is a bit too specific of a scenario to include here. It would fit better in the devcontainer.json you use to spin up your dev container.

PR #1270, opened 2024-12-26
[anaconda] - tornado - GHSA-8w49-h785-mj3c - patch to remove vuln - v6.4.1 to v6.4.2

Kaniska244:
Dev Container Name

- Anaconda

Package Name

- Tornado

Changelog

- introduced tornado version v6.4.2 in apply_security_patches.sh for tornado so that v6.4.1 isn't installed by default.

- updated tests to verify that new image contains updated v6.4.2 for tornado

Checklist

- Checked that applied changes work as expected


ddoyle2017 (approved):
LGTM

Issue #90, opened 2022-08-30
[Tracking] Image support, deprecation / EOL dates


samruddhikhandale:
The following are release dates and EOL / end of support dates for languages / OS's with pre-built images. After runtime or OS support for a given image has ended, we will stop building images for that version. We also typically only release images for "LTS" or "stable" releases of runtimes/OS's. Others can be built from the Dockerfiles available for each image in the source code in this repository.

Releases typically happen monthly, in the final week of one month or first week of the next, and include the most recent changes from the upstream base image or runtime version, including security updates (e.g. the latest OS updates and the latest point release of a runtime or platform). We may update earlier if a critical issue or bug is discovered.

Dev container images are also versioned independent of the runtime. To avoid unexpected breaks, we recommend pinning to the latest major version of an image release to pick up security and feature updates without breaking changes. This image version number is a prefix to the variant or runtime version, e.g. `mcr.microsoft.com/devcontainers/typescript-node:0-14` where `0` is the major version number of the dev container image.

To make pinning to a version easier if you wish to do so, recent versions of images include a `devcontainer-info` command that will output information about the image in use including version information and a link to a file with additional information on contents.

> **Note**: In some cases, we may have to end support earlier than the stated date below if there are fundamental OS requirements that force it or if the primary extension(s) supporting it have dropped support. Our expectation is that this should be rare given typical GA to EOL timelines and is something we strive to avoid.

| Language  / OS | Version | Release | EOL / End of Support | Image |
|-----------|---------|---------|-----|-----|
| <s>[Alpine](https://alpinelinux.org/releases/) | <s>3.13 |<s> Jan 14th, 2021 | <s>Nov 1st, 2022  | <s>[`mcr.microsoft.com/devcontainers/base:alpine-3.13`](https://github.com/devcontainers/images/tree/main/src/base-alpine) 
| <s>[Alpine](https://alpinelinux.org/releases/) | <s>3.14 | <s>June 15th, 2021 | <s>May 1st, 2023  |  <s>[`mcr.microsoft.com/devcontainers/base:alpine-3.14`](https://github.com/devcontainers/images/tree/main/src/base-alpine) |
| <s>[Alpine](https://alpinelinux.org/releases/) | <s>3.15 | <s>Nov 24th, 2021 | <s>Nov 1st, 2023  | <s>[`mcr.microsoft.com/devcontainers/base:alpine-3.15`](https://github.com/devcontainers/images/tree/main/src/base-alpine) |
[...]

PR #1276, opened 2024-12-31
Support Ruby 3.4 

Kaniska244:
**Ref**: #90 

**Devcontainer name**

Ruby

**Description**

This pr adds 3.4 latest ruby version to the supported versions list in ruby image.

**Changelog**

- Changes are there in manifest.json and readme.md file for ruby image to add support to latest 3.4 ruby version.

**Checklist**

- Checked that applied changed work as expected


NelcifranSofto:
wow Finally !!!!

ddoyle2017 (changes requested):
Changes look good! However, the [Dockerfile](https://github.com/devcontainers/images/blob/8ab31b257574d3ef183293ccf15010f6e6e7db57/src/ruby/.devcontainer/Dockerfile#L1) also needs to be updated with a reference to 3.4

src/ruby/README.md
```diff
@@ -23,6 +23,7 @@ You can directly reference pre-built versions of `Dockerfile` by using the `imag
 
 - `mcr.microsoft.com/devcontainers/ruby`     (latest)
 - `mcr.microsoft.com/devcontainers/ruby:3`   (or `3-bookworm`, `3-bullseye` to pin to an OS version)
+- `mcr.microsoft.com/devcontainers/ruby:3.3` (or `3.4-bookworm`, `3.4-bullseye` to pin to an OS version)
```
ddoyle2017, line 26: I think there's a typo here. `mcr.microsoft.com/devcontainers/ruby:3.3` should be `mcr.microsoft.com/devcontainers/ruby:3.4`

ddoyle2017 (changes requested):
Thanks for updating the `Dockerfile`! There's a small error that needs to be fixed and I think we're good to go

src/ruby/README.md
Kaniska244: Hi @ddoyle2017 ,

Thanks a lot for pointing that one out. I have corrected the same. Would you kindly review once again.

With Regards,
Kaniska

cgrothaus:
The docker image for 3.4 is not yet published (https://mcr.microsoft.com/en-us/artifact/mar/devcontainers/ruby/tags only has `dev-3.4`). When will this happen?
(Sorry if I am asking questions that would better be asked somewhere else – I don't know how this is all connected).

ashkulz:
@Kaniska244 I can see `dev-3.4` tag being published but not `3.4` -- do you know when it'll get published?

PR #1279, opened 2025-01-02
Create SECURITY.md

psyberjunkie:
Suggest MFA for all participants


ddoyle2017:
Thanks for contributing! Can you provide some more context behind your changes? I see you cited MFA here, however the `.md` file added doesn't seem to mention it at all

PR #1281, opened 2025-01-08
Revert "Automated update for image history"

Mohammed-Alanazisa:
Reverts devcontainers/images#1262


ddoyle2017:
Thanks for the contribution! However, we need to keep an updated `.md` file for each version bump, so reverting #1262 would cause some issues. I'll close this PR for now, however feel free to open a new one with more context explaining your change and we can discuss it!

PR #1287, opened 2025-01-22
[universal] - Issue universal config change for non-root default codespace user and installing google chrome browser reuse sandbox to run puppeteer cli in universal image

Kaniska244:
**Ref**: https://github.com/devcontainers/internal/issues/247 
               https://github.com/devcontainers/internal/issues/249 

**Dev container name:**

Universal

**Description:**

We can determine UID of host machine via id -u & id -g and pass it to docker while building the image. Modified devcontainer.json as below resolved the issue.

**Changelog:**

For Universal UID permission issue

- Updated GitHub workflows to determine UID of host machine via id -u & id -g and pass it to GITHUB_ENV & updated devcontainer.json file to read the UID & GID from localEnv.

For puppeteer library issue

- Updated setup-user local feature to install chrome & also changed the devcontainer.json to populate a specific env variable in the remoteEnv, required for puppeteer library.

**Checklist:**

- Checked that applied changes work as expected.

PR #1315, opened 2025-02-24
Revert "[universal] - Issue universal config change for non-root default codespace user and installing google chrome browser reuse sandbox to run puppeteer cli in universal image"

Kaniska244:
**Descripotion:** Reverts devcontainers/images#1287 partially. Would still need to keep the fix done for UID & GID issue (https://github.com/devcontainers/internal/issues/247) to ensure oryx-install-dotnet-2.1 test step is successful. Also changes done in the image size check test as suggested during review of #1287 . Need to keep that as well to ensure size check test doesn’t fail unnecessarily, checking the updated image which will not be released.

**Changelog:** 

- Change done in ./src/universal/test-project/test.sh to add ownership for codespace user to ensure oryx-install-dotnet-2.1 test step is successful. Also removed run-puppeteer test step as discussed.
- Changes done in ./.github/actions/smoke-test/check-image-size.sh to change the image size check logic so that the correct image is checked.
- Changes done in ./.github/actions/smoke-test/build.sh to change the image size check logic so that the correct image is checked.
-  Changes done in ./.github/actions/smoke-test/test.sh to change the image size check logic so that the correct image is checked.

### DIFF
--- a/.github/actions/smoke-test/build.sh
+++ b/.github/actions/smoke-test/build.sh
@@ -7,6 +7,10 @@ export DOCKER_BUILDKIT=1
 echo "(*) Installing @devcontainer/cli"
 npm install -g @devcontainers/cli
 
-echo "(*) Building image - ${IMAGE}"
 id_label="test-container=${IMAGE}"
+id_image="${IMAGE}-test-image"
+echo "(*) Building image - ${IMAGE}"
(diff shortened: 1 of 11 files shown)


ddoyle2017 (approved):
Thanks for following up our discussion with this!

For context: introducing Chrome for puppeteer to use during the test step is causing image size issues, so we're temporarily disabling the test + rolling back adding Chrome

PR #1325, opened 2025-03-02
Update push-dev.yml

Dondiago:
/feed subscribe https://status.openai.com/slack.rss


Kaniska244:
Hi @Dondiago ,

Looks like this was created by mistake, so closing the same. Kindly let me in case of any concern.

With Regards,
Kaniska

PR #1345, opened 2025-03-21
Support: Java 24

m1ngyuan:
In the case of supporting all LTS branches, it is possible to consider maintaining a latest NON-LTS version to facilitate users' attempts at new JDK features


src/java/manifest.json
```diff
@@ -14,6 +16,10 @@
 		"parent": "base-debian",
 		"idMismatch": "true",
 		"variantBuildArgs": {
+			"24-bookworm": {
+				"TARGET_JAVA_VERSION": "24",
```
Mathiyarasy, line 20: Devcontainer java image docker file is using microsoft openjdk and version 24 is not yet available.
Source: https://learn.microsoft.com/en-us/java/openjdk/.

m1ngyuan: [JDK 25](https://learn.microsoft.com/en-us/java/openjdk/download#openjdk-2500-lts) is now available

perry2of5:
Java 25 has been out for nearly a month now and the microsoft open JDK build has been out for more than 2 weeks. It would be nice to have the Java 25 dev containers released as well.

Is there anything I can do to help speed this along?

PR #1404, opened 2025-05-16
[universal] - Image change ubuntu focal EOL

Kaniska244:
**Ref:** https://github.com/devcontainers/images/issues/90

**Devcontainer Image:**

- Universal

**Description of changes:**

- Changes corresponding to Ubuntu focal (20.04) EOL. Upgraded the base image for universal image to Ubuntu noble (24.04)

**Changelog:**

- Change in Dockerfile to update the base image to noble along with changes to upgrade some system libraries & bringing them as one layer to reduce size & also removed puppeteer related librarries as they are not needed anymore as puppeteer isn't supported by universal image currently. Removed the default `ubuntu` user from the base image so that the `codespace` user creation is successful.

- Small change in test-utils.sh to update system library references. 

Checklist:

- Checked that applied changes work as expected

### DIFF
--- a/src/universal/.devcontainer/Dockerfile
+++ b/src/universal/.devcontainer/Dockerfile
@@ -3,18 +3,23 @@
 # Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
 #-------------------------------------------------------------------------------------------------------------
 
-FROM ubuntu:focal
+FROM ubuntu:noble
+#FROM buildpack-deps:noble-curl
+
+RUN if id "ubuntu" &>/dev/null; then \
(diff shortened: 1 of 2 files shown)


AlvaroRausell:
@Kaniska244 Instead of installing libssl1, can you try consuming the `install.sh` script from the oryx feature? https://github.com/devcontainers/features/blob/main/src/oryx/install.sh

src/universal/manifest.json
```diff
@@ -4,14 +4,14 @@
 		"latest": true,
 		"rootDistro": "debian",
 		"tags": [
-			"universal:${VERSION}-focal",
+			"universal:${VERSION}-noble",
```
eljog, line 7: If we don't update version to 3.0.0 above, will the dev images be released with correct tagging?
If so, update it now?

eljog (dismissed):

eljog (approved):

PR #267, opened 2022-12-13
Universal: Update versions of languages/SDKs

samruddhikhandale:
- https://github.com/nodejs/release#release-schedule
- https://www.php.net/supported-versions.php

PR #1428, opened 2025-05-30
[anaconda] - Fix to stick to transformers version 4.49.0

Kaniska244:
Ref# [#267](https://github.com/devcontainers/internal/issues/267)

**Description:** This change is to create the ability to pin extra python packages to specific versions provided as input instead of looking for higher version.

**Changelog:** The following changes have been made.
- Change done in `images/src/anaconda/.devcontainer/apply_security_patches.sh` introduce an array `pin_to_required_version` which should consist of package names for which the install should stick to the given version in `vulnerable_packages` array instead of looking for higher version.
- Version bump.

**Checklist:**
- [x] All checks are passed.

### DIFF
--- a/src/anaconda/.devcontainer/apply_security_patches.sh
+++ b/src/anaconda/.devcontainer/apply_security_patches.sh
@@ -3,7 +3,7 @@
 # vulnerabilities:
 # werkzeug - [GHSA-f9vj-2wh5-fj8j]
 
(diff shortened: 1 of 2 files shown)


AlvaroRausell (approved):
Not in the scope of this PR, but I think the right way to do this would be to use requirements.txt instead of this

PR #1555, opened 2025-09-17
Bump Yarn 1.22.22 -> 4.9.4 - Remove sha1 hashes 

sam-byng:
See ADO task [TASK 2576673](https://devdiv.visualstudio.com/OnlineServices/_workitems/edit/2576673)

NOTE: SHA-1 is not allowed for any purpose, visit [Microsoft.Security.Cryptography.10021](https://aka.ms/Cryptography10021) for details.

Therefore bumped to latest yarn javascript package manager to avoid sha1 hashing of package binaries.

This adds the yarn berry config / package files: .yarn/ project-local workspace and .yarnrc.yml berry config
Note: Here we bootstrap yarn in devcontainer postCreate. WE also download yarn 4.9.4 locally and refer to it from package.json. Just in case devcontainers not used.


sam-byng:
@microsoft-github-policy-service agree company="Microsoft"

AlvaroRausell (approved):

.devcontainer/devcontainer.json
```diff
@@ -4,7 +4,7 @@
         "ghcr.io/devcontainers/features/docker-in-docker:2": {},
         "ghcr.io/devcontainers/features/azure-cli:1": {}
     },
-    "postCreateCommand": "npm install -g @devcontainers/cli",
+    "postCreateCommand": "corepack enable && corepack prepare yarn@4.9.4 --activate && npm install -g @devcontainers/cli",
```
AlvaroRausell: Actually, could we not install the yarn feature and set it to 4.9.4 instead?

AlvaroRausell (changes requested):

.devcontainer/devcontainer.json
sam-byng: AFAICT there's know yarn feature in https://containers.dev/features. Have tested the [node feature ](https://github.com/devcontainers/features/tree/main/src/node) but it has the old yarn 1.22.22 and not the up-level version, doesn't allow you to set yarn version either.

AlvaroRausell: Hmm, we should create an issue about that

AlvaroRausell:
@sam-byng I approved, but can you create an issue against the features repo, so that we can allow for newer versions of yarn please?

PR #1582, opened 2025-10-14
Update Packages via Python due to Conda Channel Deprecation (Transitioning from v24.5.0 to v25.9.0)

sireeshajonnalagadda:
Devcontainer Image:

Universal
Description of changes:

Updating the packages (urllib3, requests and setuptools ) through Python as the Conda channel is deprecating these packages.
The packages are not found because the Conda version is transitioning from version 24.5.0 to 25.9.0.
Changelog:

Change in patch-conda/install.sh 
Checklist:

All checks are passed.


sireeshajonnalagadda:
another PR created

PR #1586, opened 2025-10-14
[Universal] Use the default conda channel

Mathiyarasy:
**Fixes**: https://github.com/devcontainers/internal/issues/314

**Issue :**
Conda is no longer adding 'defaults' to channel list implicitly from version 25.9.
https://github.com/conda/conda/issues/14476

**Solution :**
The conda channels must be explicitly specified

### DIFF
--- a/src/dotnet/README.md
+++ b/src/dotnet/README.md
@@ -9,10 +9,10 @@
 | *Categories* | Core, Languages |
 | *Image type* | Dockerfile |
 | *Published images* | mcr.microsoft.com/devcontainers/dotnet |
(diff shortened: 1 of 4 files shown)


AlvaroRausell:
@Mathiyarasy Do you also need to add this to the conda image?

v-Mathiyarasy:
> @Mathiyarasy Do you also need to add this to the conda image?

The current change only affects conda version 25.9.0. And both conda devcontainer images build are successful so far in the weekly builds. In both conda devcontainer images, conda is not installed via devcontainer features.
The anaconda devcontainer image uses continuumio/anaconda3 as its base, which includes conda version 24.9.2.
The miniconda image uses continuumio/miniconda3 as its base, where the conda version is 25.3.1.

AlvaroRausell (approved):

AlvaroRausell:
@Mathiyarasy I have updated the PR name to better match this, but approved. Will merge once CI is done

PR #1613, opened 2025-11-02
Delete src/universal directory

harem041997:
src/universal/.devcontainer/devcontainer.json


bamurtaugh:
Please let us know if there was a specific issue you were aiming to solve with this type of change. Thank you!

PR #1627, opened 2025-11-10
Alpine 3.19 EOL

sireeshajonnalagadda:
Ref [#1464 ](https://github.com/devcontainers/images/discussions/1464)


src/base-alpine/manifest.json
```diff
@@ -1,10 +1,9 @@
 {
-	"version": "1.0.2",
```
AlvaroRausell: Can you bump this to 2.0.0 please?

sireeshajonnalagadda: Bumped up the version.

PR #1652, opened 2025-11-20
Remove uninstallation of imagemagick from debian

jemsab:
The current version of the Rust Dockerfile removes the imagemagick package from the OCI image because of CVE CVE-2019-10131. 

According to Debian's security tracker https://security-tracker.debian.org/tracker/CVE-2019-10131, the vulnerability has been fixed in all Debian versions referenced in the manifest file (bullseye, bookworm and trixie). This workaround is therefore no longer necessary.

Removing the workaround will also remove a layer from the generated OCI image reducing deployment size.


abdurriq:
Thank you for your contribution @jemsab. Please agree to the CLA and we can then merge your change.

jemsab:

abdurriq (approved):

PR #1663, opened 2025-11-26
Build workflow modification

Kaniska244:
**Description of changes:** 

- Changing the trigger of dev container images build workflow trigger such that it will be triggered when a new release is published in images repository when the target branch is the `main` branch.

**Changelog:**

- Change in /.github/workflows/push.yml to change the trigger.

**Checklist:**
- [x] All checks are passed.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -3,7 +3,7 @@ name: Build and push images
 on:   
   release:
     types: [published]
-workflow_dispatch:
+  workflow_dispatch:
 
 jobs:
   build-and-push:


AlvaroRausell (approved):

.github/workflows/push.yml
```diff
@@ -1,14 +1,13 @@
 name: Build and push images
 
-on: 
-  push:
-    tags:
-    - 'v*'
+on:   
+  release:
+    types: [published]
```
AlvaroRausell, line 5: @Kaniska244 Can you add workflow dispatch too, so we can trigger it manually when needed (like now)?

v-Kaniska244: Hi @AlvaroRausell 

Added the same. As of now, this workflow doesn't get triggered with workflow_dispatch in its present form.

PR #1724, opened 2026-01-06
Add Support for Ruby 4.0.0

sireeshajonnalagadda:
Devcontainer image 
Ruby 

Adding support for Ruby 4.0 version

Changes
Updated the manifest and Dockerfile and Readme files 

Checklist
Changes works as expected

PR #1746, opened 2026-01-21
Fix for dev container images build failure

Kaniska244:
**Devcontainer Image:**

- ruby

**Description of changes:** 

- Aims to fix dev container image build [failure](https://github.com/devcontainers/images/actions/runs/21140443475). Due to malformed json updates done as part of [PR](https://github.com/devcontainers/images/pull/1724), the `vscdc/build` component isn’t able to parse the dependencies for ruby image resulting in null/ undefined error crashing the build process altogether. Additionally found another issue with the same PR which is that `4.0-bullseye` ruby image isn't available upstream but it was still added as a ruby dev container image variant.  

**Changelog:**

- Changes in manifest.json to correct the malformed json updates and removing `4.0-bullseye` variant.

**Checklist:**
- [x] All checks are passed.

### DIFF
--- a/src/ruby/README.md
+++ b/src/ruby/README.md
@@ -9,7 +9,7 @@
 | *Categories* | Core, Languages |
 | *Image type* | Dockerfile |
 | *Published images* | mcr.microsoft.com/devcontainers/ruby |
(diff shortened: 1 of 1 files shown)


abdurriq:
@Kaniska244 Could you please update the README to remove 4.0-bullseye. Also, could we add JSON validation for all the config to avoid this recurring the future?

v-Kaniska244:
> @Kaniska244 Could you please update the README to remove 4.0-bullseye. Also, could we add JSON validation for all the config to avoid this recurring the future?

Hi @abdurriq ,

I have corrected the readme file now. Indeed a json validation check would be helpful mitigating such problems. Let me look into it further and get back with another PR. Will create an issue in internal repo to track the same.

PR #1750, opened 2026-01-22
fix: remove empty markdown line which broke table rendering

cngJo:
Small markdown formatting fix. Noticed it ~2weeks ago and finally had the time to fix it :) 

Not a notable contribution but wanted to fix it nevertheless. Thank you for providing those devcontainer images and the system in general 🎉


abdurriq:
@cngJo Thanks for being proactive and fixing this!

Issue #1752, opened 2026-01-23
GPG key for Yarn has expired,


mfoote-bx:
The GPG key for Yarn has expired, again, it seems like this happened before:
https://github.com/devcontainers/images/issues/370

Example from mcr.microsoft.com/devcontainers/python:dev-3.14-bookworm running apt update
#18 0.335 Err:4 https://dl.yarnpkg.com/debian stable InRelease
#18 0.335   The following signatures were invalid: EXPKEYSIG 23E7166788B63E1E Yarn Packaging 
#18 1.061 Reading package lists...
#18 1.500 W: GPG error: https://dl.yarnpkg.com/debian stable InRelease: The following signatures were invalid: EXPKEYSIG 23E7166788B63E1E Yarn Packaging 
 
There is an issue in Yarnpkg about it, I guess this would need to be fixed before the images are recreated
https://github.com/yarnpkg/yarn/issues/9216

If you remove yarn you can run apt-get. It seems using v1 or Yarn is not so advisable these days?
https://github.com/devcontainers/images/blob/main/src/python/manifest.json#L92-L95

Daniel15:
Grab the new key and you should be OK

```
gpg --keyserver keyserver.ubuntu.com --recv-keys 23E7166788B63E1E
```

JartanFTW:
Upstream (https://github.com/yarnpkg/yarn/issues/9216) documents the Yarn repository GPG key expiration issue. Although the key’s expiry has now been [extended upstream](https://github.com/yarnpkg/yarn/issues/9216#issuecomment-3792637538), the current devcontainer images still embed the old expired key, so apt-get update continues to fail. A new image rebuild that refreshes the keyring is needed for builds to succeed.

JartanFTW:
@Kaniska244 You seem like an active maintainer of this repository. Could you please trigger a rebuild to fix this issue?
Er actually on closer inspection, maybe I should be tagging @abdurriq instead.

flazzarini:
For now I have fixed the issue using the following command during the dev-container build process. I am using `python:3.13-bookworm` 

```
# -----------------------------------------------------------------------------
#  Fix yarn signature issue, open bug with microsoft images
#   https://github.com/devcontainers/images/issues/1752
# -----------------------------------------------------------------------------
curl -fsSL https://dl.yarnpkg.com/debian/pubkey.gpg \
  | sudo gpg \
        --batch \
        --yes \
        --dearmor \
[...]

PR #1804, opened 2026-03-17
Fix Yarn GPG key issue in Jekyll Dockerfile

Rabadash8820:
My Jekyll-based devcontainers have been failing to build for a month or two (see this [Codespace prebuild log](https://github.com/Rabadash8820/Dansite/actions/runs/23179269576/job/67348504809) for example) due to the Yarn Classic GPG error described in #1752. I'm not really sure why the Jekyll devcontainer is adding the Node feature at all, but if it's going to then I think it should at least remove Yarn Classic since that version is no longer being maintained. This PR accomplishes that by completely removing the Yarn apt repository from the Jekyll devcontainer (inspired by the .NET Aspire team's related PR: dotnet/aspire-devcontainer#7).

I'm not positive if this PR will solve the issue as I wasn't sure how to test it, but hopefully this gets the ball rolling. 👍

### DIFF
--- a/src/jekyll/.devcontainer/Dockerfile
+++ b/src/jekyll/.devcontainer/Dockerfile
@@ -11,6 +11,12 @@ ENV LANG=en_US.UTF-8 \
     LANG=en_US.UTF-8 \
     LANGUAGE=en_US
 
+# Fix Yarn GPG key issue - remove the problematic Yarn repository
+# that has an expired/rotated signing key causing apt-get update to fail
+# See: https://github.com/devcontainers/images/issues/1752
+# Inspired by this PR: https://github.com/dotnet/aspire-devcontainer/pull/7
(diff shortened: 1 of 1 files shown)


Kaniska244:
Hi @Rabadash8820 

Thank you for your contribution. This problem has been fixed already. As per the log you have shared, the base image used is `mcr.microsoft.com/devcontainers/jekyll:2.2.5-3.4-bookworm` which is an older image without the fix. Request you to use the latest `mcr.microsoft.com/devcontainers/jekyll:2.2.7-3.4-bookworm` or any other version agnostic editions which also received the latest updates.

Rabadash8820:
Awesome, yes I can confirm that the devcontainer builds with `jekyll:2.2.7-3.4-bookworm`. Sorry for the noise, thank you!

PR #1820, opened 2026-03-30
[universal] - Update toolings and vulnerability fix

Kaniska244:
**Changelog:** The following files are changed.

- Updated the versions of python and ruby
- Fix for vulnerability issue [GHSA-r6ph-v2qm-q3c2](https://github.com/advisories/GHSA-r6ph-v2qm-q3c2) for universal image. 
- Tests modified.
- Version bump.
- JSON linter started failing with ESLint error due to update in super-linter@v8 action to ESLint `9.39.2+` where `jsonc-eslint-parser` dropped its default export, breaking the old import `jsoncParser` from `jsonc-eslint-parser`. Using FlatCompat as a workaround then caused a circular structure error. The native flat config avoids both issues. Removed `@eslint/eslintrc (FlatCompat)` and `jsonc-eslint-parser` imports as both are no longer needed. Added direct import of `eslint-plugin-jsonc`, which natively supports ESLint flat config.

**Checklist:**
- [x] All checks are passed.

### DIFF
--- a/src/universal/.devcontainer/local-features/patch-conda/install.sh
+++ b/src/universal/.devcontainer/local-features/patch-conda/install.sh
@@ -51,10 +51,10 @@ sudo_if /opt/conda/bin/python3 -m pip install --upgrade pip
 # Temporary: Upgrade python packages due to security vulnerabilities
 # They are installed by the conda feature and Conda distribution does not have the patches
 
-# https://github.com/advisories/GHSA-79v4-65xg-pq4g
-update_python_package /opt/conda/bin/python3 cryptography "44.0.1"
+# https://github.com/advisories/GHSA-r6ph-v2qm-q3c2
+update_conda_package pyopenssl "26.0.0"
 
-update_conda_package pyopenssl "25.0.0"
+update_python_package /opt/conda/bin/python3 cryptography "46.0.5"
 
 # https://github.com/advisories/GHSA-9hjg-9r4m-mvj7
 update_conda_package requests "2.32.4"
--- a/src/universal/.devcontainer/local-features/patch-conda/install.sh
+++ b/src/universal/.devcontainer/local-features/patch-conda/install.sh
@@ -54,7 +54,7 @@ sudo_if /opt/conda/bin/python3 -m pip install --upgrade pip
 # https://github.com/advisories/GHSA-r6ph-v2qm-q3c2
 update_conda_package pyopenssl "26.0.0"
 
-update_python_package /opt/conda/bin/python3 cryptography "46.0.5"
+update_conda_package cryptography "46.0.5"
 
 # https://github.com/advisories/GHSA-9hjg-9r4m-mvj7
 update_conda_package requests "2.32.4"
--- a/.github/linters/eslint.config.mjs
+++ b/.github/linters/eslint.config.mjs
@@ -2,7 +2,7 @@ import { defineConfig } from "eslint/config";
 import { FlatCompat } from "@eslint/eslintrc";
 import js from "@eslint/js";
 import globals from "globals";
-import jsoncParser from "jsonc-eslint-parser";
+import * as jsoncParser from "jsonc-eslint-parser";
 
 const compat = new FlatCompat();
 
--- a/.github/linters/eslint.config.mjs
+++ b/.github/linters/eslint.config.mjs
@@ -1,10 +1,7 @@
 import { defineConfig } from "eslint/config";
-import { FlatCompat } from "@eslint/eslintrc";
 import js from "@eslint/js";
 import globals from "globals";
-import * as jsoncParser from "jsonc-eslint-parser";
-
-const compat = new FlatCompat();
+import jsoncPlugin from "eslint-plugin-jsonc";
 
 export default defineConfig([
   js.configs.recommended,
@@ -44,12 +41,5 @@ export default defineConfig([
     },
   },
 
-  ...compat.extends("plugin:jsonc/recommended-with-jsonc"),
-  {
-    files: ["**/*.json"],
-    languageOptions: {
-      parser: jsoncParser,
-      parserOptions: { jsonSyntax: "JSONC" },
-    },
-  },
+  ...jsoncPlugin.configs["flat/recommended-with-jsonc"],
 ]);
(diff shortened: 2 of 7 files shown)


abdurriq (approved):

abdurriq (dismissed):

src/universal/.devcontainer/local-features/patch-conda/install.sh
```diff
@@ -51,10 +51,10 @@ sudo_if /opt/conda/bin/python3 -m pip install --upgrade pip
 # Temporary: Upgrade python packages due to security vulnerabilities
 # They are installed by the conda feature and Conda distribution does not have the patches
 
-# https://github.com/advisories/GHSA-79v4-65xg-pq4g
-update_python_package /opt/conda/bin/python3 cryptography "44.0.1"
+# https://github.com/advisories/GHSA-r6ph-v2qm-q3c2
+update_conda_package pyopenssl "26.0.0"
 
-update_conda_package pyopenssl "25.0.0"
+update_python_package /opt/conda/bin/python3 cryptography "46.0.5"
```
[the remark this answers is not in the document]
abdurriq: @Kaniska244 Is this accurate? Should we only be using update_conda_package for conda patches?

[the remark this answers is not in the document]
Kaniska244: Hi @abdurriq ,

The `update_python_package` is used to carry on with the installation with python wheel in case if the specified package version not found in the conda channel configured by default. The `cryptography` package was using the `update_python_package` function even in the previous version of the script, so I didn't change it in the first place. However as the conda installation is always the better and cleaner approach and this specific version of `cryptography` seems to be available in the conda channel, so i have changed it. Kindly let me know in case of any further concern.

src/universal/.devcontainer/local-features/patch-conda/install.sh
abdurriq, line 66: Happy to approve, but could we look into removing this going forward so that the file only updates conda packages, not pip?

abdurriq (approved):

src/universal/.devcontainer/local-features/patch-conda/install.sh
Kaniska244: Hi @abdurriq 

The problem happens is that if we receive a vulnerability issue with one of these python libraries asking to upgrade to a specific version and the specified version not available in conda yet, we resort to installation of the upgraded library using pip. However, going forward will check for more options such as using different onda channel to solve the problem before using python wheels using pip as last resort.

Kaniska244: Hi @abdurriq,

To add more details on the `tqdm` python library installation, it has direct dependency on the conda environment and this specific version (4.66.3) doesn't seem to be available in any conda channel. Updating to higher available version failing with the following error:

```
RemoveError: 'tqdm' is a dependency of conda and cannot be removed from
conda's operating environment.
```

Below details from copilot on this:

```
# Plan: Fix tqdm installation in patch-conda

## TL;DR
tqdm is a protected conda dependency and can't be uninstalled/replaced via `conda install` or the current `update_python_package` (which does `pip uninstall` first). Fix by using `pip install --force-reinstall --no-deps` to upgrade tqdm in-place without triggering conda's dependency protection.

## Problem
1. `tqdm 4.66.3` is not available on any conda channel (defaults, conda-forge, anaconda, bioconda)
2. `tqdm` is a runtime dependency of conda itself — conda blocks its removal
3. `conda install -c conda-forge tqdm=4.66.4` also fails with `RemoveError: 'tqdm' is a dependency of conda`
4. The current `update_python_package` function does `pip uninstall --yes` first, which will either fail or corrupt conda's environment

## Steps

### Phase 1: Add a new helper function for conda-protected packages

1. Add a new function `update_conda_protected_package` in `install.sh` that uses `pip install --force-reinstall --no-deps --no-cache-dir` to upgrade packages in-place without uninstalling. This avoids triggering conda's dependency protection.
   - Place it after the existing `update_conda_package` function (around line 50)
   - Function signature: `update_conda_protected_package PYTHON_PATH PACKAGE VERSION`
   - Implementation: `sudo_if "$PYTHON_PATH -m pip install --force-reinstall --no-deps --no-cache-dir $PACKAGE==$VERSION"`
   - Then verify with: `sudo_if "$PYTHON_PATH -m pip show --no-python-version-warning $PACKAGE"`

2. Replace the tqdm `update_python_package` call (line 66) with `update_conda_protected_package`:
   - Before: `update_python_package /opt/conda/bin/python3 tqdm "4.66.3"`
   - After: `update_conda_protected_package /opt/conda/bin/python3 tqdm "4.66.3"`

### Phase 2: Version bump in manifest.json (per repo copilot-instructions)

3. Bump patch version in `src/universal/manifest.json` from `6.0.1` to `6.0.2` (security/bug fix = patch bump)

## Relevant files
- `src/universal/.devcontainer/local-features/patch-conda/install.sh` — add new helper function, update tqdm install call
- `src/universal/manifest.json` — bump version from 6.0.1 to 6.0.2

## Verification
1. Run `sudo bash /workspaces/images/src/universal/.devcontainer/local-features/patch-conda/install.sh` in the dev container and confirm tqdm installs without errors
2. Verify tqdm version after install: `/opt/conda/bin/python3 -c "import tqdm; print(tqdm.__version__)"` should output `4.66.3`
3. Verify conda still works: `conda --version` should succeed (not broken by tqdm replacement)
4. Verify `manifest.json` version is bumped

## Decisions
- Use `--force-reinstall --no-deps` instead of uninstall+install to bypass conda's dependency protection
- Keep version 4.66.3 (available on PyPI, addresses GHSA-g7vv-2v7x-gj9p)
- Keep using pip (not conda) since the target version isn't on any conda channel
- New helper function is preferred over inlining, since other conda-protected packages (e.g., setuptools, certifi) could need this approach in the future
- `--no-deps` is important to avoid pulling in pip-resolved dependencies that conflict with the conda environment

## Further Considerations
1. **Should we bump tqdm to 4.67.x instead?** The fix version 4.66.3 patches the advisory. A newer version would also work but introduces more change. Recommend keeping 4.66.3 for minimal diff.
2. **Could other packages in this script hit the same issue?** If conda ever protects setuptools, urllib3, etc. the same way, they'd need the same treatment. The new helper function makes this easy to switch.

```

abdurriq: I see, thanks for clarifying.

PR #1903, opened 2026-06-23
[smoke-test] fix a flake: retry tag validation to tolerate Docker Hub rate limiting

V-Subhankar-infy:
## Summary
Smoke-test tag validation failed intermittently with false `Invalid variants`
errors (e.g. `3.13-trixie`) when anonymous Docker Hub calls got rate limited.

## Root cause
`validate-tags.sh` ran a single `docker manifest inspect` per tag and treated
any non-zero exit — including HTTP 429 rate limiting — as "tag does not exist",
so one throttled call failed the whole job despite the tag existing upstream.

## Changes (`check_image_exists()` only)
- Retry the registry query up to 3× with a 10s pause before marking a tag invalid.
- On failure, flag the one unambiguous transient cause — rate limiting (HTTP 429);
  everything else keeps the original `✗ Invalid - tag does not exist` message.
- Echo the raw registry response for debugging.

## Preserved
Fail-closed: genuinely missing tags still `exit 1`, and the default message is unchanged and displays "invalid-tag"

## Testing
`bash -n` passes; classification self-test confirms 429 → rate-limit message and
all else → original message;


.github/actions/smoke-test/validate-tags.sh
```diff
@@ -29,19 +29,41 @@ VARIANTS=$(jq -r '.variants[]?' "$MANIFEST_FILE" 2>/dev/null || true)
...
+        # must not be misreported as "tag does not exist".
+        if echo "$output" | grep -qiE 'manifest unknown|manifest for .* not found|no such manifest'; then
+            echo "    ✗ Invalid - tag does not exist"
+            return 1
+        fi
+
+        echo "    ! Transient error (attempt ${attempt}/${max_attempts}): $(echo "$output" | tr '\n' ' ' | cut -c1-200)"
+        attempt=$(( attempt + 1 ))
+        if (( attempt <= max_attempts )); then
+            sleep $(( (attempt - 1) * 5 ))
```
Kaniska244: It would be better if we extend the delay a bit further given that we are doing this avoid cascading effect of rate limiting in DockerHub, something like `sleep $(( (attempt - 1) * 10 ))`.

.github/actions/smoke-test/validate-tags.sh
```diff
@@ -117,5 +147,21 @@ if [[ ${#INVALID_TAGS[@]} -gt 0 ]]; then
     exit 1
 fi
 
+if [[ ${#UNVERIFIED_TAGS[@]} -gt 0 ]]; then
+    echo ""
+    echo "WARNING: Could not verify $(if [[ -n "$VARIANTS" ]]; then echo "variants"; else echo "base images"; fi) (${#UNVERIFIED_TAGS[@]}) due to transient errors (e.g. Docker Hub rate limiting):"
```
Kaniska244: This I am not sure, if would wise to proceed like this even if its because of transient failure. What if there's any outage in DockerHub and the error message doesn't contain strings used as filter inside `check_image_exists()` function. In such cases failing fast would be a safer and better strategy in my humble opinion. The retry is indeed a good addition which could be taken forward but skipping validation errors altogether may not be the wisest move. Let's get the maintainer's opinion on the same.

Kaniska244 (dismissed):

.github/actions/smoke-test/validate-tags.sh
V-Subhankar-infy: ok, sure will update it to wait for longer time.

.github/actions/smoke-test/validate-tags.sh
V-Subhankar-infy: I have considered all your points and made the following changes,
1) Removed the complex string match that depends on dockerhub response data type.
2) Removed Complex code to handle multiple cases and sticked to simple sleep to wait for 10 seconds and retry 2 more times before stopping the build.
3) Only when the docker-hub sends a definitive response code of 429 that only means rate limiting we will log that onto console.
4) The code only updates the log displayed in case of failure, the core logic is untouched & the error validation is unaffected.

.github/actions/smoke-test/validate-tags.sh
```diff
@@ -31,17 +31,44 @@ INVALID_TAGS=()
...
+    while (( attempt <= max_attempts )); do
+        if output=$(docker manifest inspect "$image_tag" 2>&1); then
+            echo "    ✓ Valid"
+            return 0
+        fi
+
+        echo "    ! Attempt ${attempt}/${max_attempts} could not verify tag"
+        attempt=$(( attempt + 1 ))
+        if (( attempt <= max_attempts )); then
+            sleep 10
```
Kaniska244: I would still suggest you use the earlier modification I proposed on the on top of the previous change. The waiting time should proportionally increase in case of multiple retries which you had implemented earlier. I only suggested to increase the gap.

Kaniska244 (dismissed):

.github/actions/smoke-test/validate-tags.sh
V-Subhankar-infy: ok, updated now.

PR #1941, opened 2026-07-29
Use GitHub App token instead of PAT in release workflows

abdurriq:
### What
Replaces the fine-grained PAT (`secrets.PAT`) in the release-automation workflows with a short-lived GitHub App installation token minted at runtime via `actions/create-github-app-token`.

Affected workflows:
- `.github/workflows/release-pr.yml` (monthly release PR: push manifest updates + create PR)
- `.github/workflows/version-history.yml` (image version-history PR)

### Why
Microsoft open source now caps PAT lifetime at 8 days (down from 3 months), which would force an 8-day rotation cadence. The App private key has no forced expiry, and the installation token it mints is automatically short-lived (~1h) and rotated every run. A PR opened under the App identity still triggers downstream required checks (unlike the built-in `GITHUB_TOKEN`). See devcontainers/internal#343.

### How it works
Each job now:
1. Mints a token with `actions/create-github-app-token@v2` using `vars.DEVCONTAINERS_REPO_AUTOMATION_ID` (org variable) and `secrets.DEVCONTAINERS_REPO_AUTOMATION_PRIVATE_KEY` (org secret, scoped to this repo).
2. Uses that token for `actions/checkout` (so `git push` is attributed to the App) and for the `git push` / `gh api .../pulls` steps in place of `secrets.PAT`.

### Prerequisites (already completed)
- `devcontainers-repos-automation` GitHub App registered on the org and installed on `devcontainers/images` (Contents R/W, Pull requests R/W, Issues R/W, Metadata R).
- Org variable `DEVCONTAINERS_REPO_AUTOMATION_ID` and org secret `DEVCONTAINERS_REPO_AUTOMATION_PRIVATE_KEY` added, scoped to this repo.

### Rotation
Key rotation is documented in the runbook (see devcontainers/internal). Rotation is zero-downtime and does not require the old key.

### Testing

Validated the App-token setup in `devcontainers/images` before merge using two throwaway, push-triggered workflows (each on a temporary branch, auto-cleaned up — no changes to `main`):

1. **Token mint + read + `contents: write`** — [run 30457758847](https://github.com/devcontainers/images/actions/runs/30457758847)
   - `create-github-app-token` minted an installation token scoped to `devcontainers/images`.
   - Confirmed read access (`gh api /repos/...`).
   - Confirmed `contents: write` by creating and deleting a temporary ref.

2. **`pull-requests: write` (PR creation via the App token)** — [run 30538125085](https://github.com/devcontainers/images/actions/runs/30538125085)
   - Opened a draft PR via `gh api POST /pulls` (the exact step both `release-pr.yml` and `version-history.yml` rely on), then auto-closed it and deleted the temporary branches.

Together these exercise every privileged operation these workflows perform (token mint, `checkout`, `git push`, `gh api .../pulls`). All test branches and the temporary test PR were closed/deleted; no artifacts remain.

> Note: `version-history.yml` is a reusable workflow referenced by `push.yml` / `push-dev.yml` as `...version-history.yml@main`, so its change takes effect on the next `push.yml` run after merge.

### Notes / follow-ups
- `secrets.PAT` can be removed from the repo once this is merged and a release run is verified green.
- Consider pinning `actions/create-github-app-token` to a full commit SHA if the repo adopts action pinning.

### DIFF
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -9,7 +9,7 @@ on:
         required: true
 
 permissions:
-  contents: write
+  contents: read
 
 
 jobs:
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -7,7 +7,7 @@ on:
         type: string
         description: New version, in the form v0.4.x
         required: true
-        
+
 permissions:
   contents: write
 
@@ -19,18 +19,26 @@ jobs:
       target_branch: rel/${{github.event.inputs.version}}
       version: ${{github.event.inputs.version}}
     steps:
+      - name: Generate a token
+        id: app-token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ vars.DEVCONTAINERS_REPO_AUTOMATION_ID }}
+          private-key: ${{ secrets.DEVCONTAINERS_REPO_AUTOMATION_PRIVATE_KEY }}
       - uses: actions/checkout@v6
+        with:
+          token: ${{ steps.app-token.outputs.token }}
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
           node-version: 20
       - name: Install devcontainers CLI
         run: npm i -g @devcontainers/cli
-      - name: Update manifest files 
+      - name: Update manifest files
         run: ./build/prepare-release.sh
       - name: Push manifest updates
         env:
-          GITHUB_TOKEN: ${{ secrets.PAT }}
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
         run: |
           git config --global user.email github-actions@github.com
           git config --global user.name github-actions
@@ -40,7 +48,7 @@ jobs:
           git push origin $target_branch
       - name: Create pull request
         env:
-          GITHUB_TOKEN: ${{ secrets.PAT }}
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
         run: |
           gh api \
                 --method POST \
@@ -50,5 +58,3 @@ jobs:
               -f body="" \
               -f head="$target_branch" \
               -f base='main'
-
- 
--- a/.github/workflows/version-history.yml
+++ b/.github/workflows/version-history.yml
@@ -31,8 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     environment: publishing
     permissions:
-      contents: write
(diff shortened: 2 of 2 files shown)


.github/workflows/release-pr.yml
```diff
@@ -7,7 +7,7 @@ on:
         type: string
         description: New version, in the form v0.4.x
         required: true
-        
+
 permissions:
   contents: write
```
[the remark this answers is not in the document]
abdurriq: Good catch — reduced to `contents: read`. All writes (checkout, git push, PR creation) now go through the App installation token, so the default `GITHUB_TOKEN` no longer needs write access. Kept `read` as a safe minimum for checkout fallback / setup-node.

.github/workflows/version-history.yml
```diff
@@ -35,86 +35,95 @@ jobs:
       pull-requests: write
```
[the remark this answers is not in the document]
abdurriq: Agreed — removed `pull-requests: write` and reduced `contents` to `read`. Push and PR creation both use the App token now, so the default token doesn't need write on either scope.

Kaniska244 (approved):
LGTM

.github/workflows/release-pr.yml
```diff
@@ -7,9 +7,9 @@ on:
         type: string
```
sam-byng, line 7: Q: Is this tested? Could you run this from branch and put results in PR description?

.github/workflows/release-pr.yml
```diff
@@ -7,9 +7,9 @@ on:
...
         description: New version, in the form v0.4.x
```
sam-byng, line 8: Suggestion: worth updating some docs - possibly in devcontainers/internal to note the change here? Perhaps just a comment to link to the relevant github app we're using (`DEVCONTAINERS_REPO_AUTOMATION_ID`)

sam-byng (changes requested):

.github/workflows/release-pr.yml
abdurriq: Yup, I've got a docs PR for this: https://github.com/devcontainers/internal/pull/344

.github/workflows/release-pr.yml
abdurriq: Yes, I was able to test it; updated the PR description too.

sam-byng (approved):
Thanks LGTM!

